PR #5, opened 2025-02-10
feat: Initial storybook setup

Etwigg:
First pass at storybook implementation and some other small changes.


shaunrfox (approved):
Looks good to me. My refactor branch will affect some of the same files, but I don't think conflicts will be too bad.

PR #8, opened 2025-02-26
Component/check box component new

Yash-CETEC:


Yash-CETEC:
@Etwigg and @shaunrfox can you please take a look at this pr?

Etwigg (approved):

Etwigg:
@Yash-CETEC 
Shaun changes how we're gonna handle the checkbox by introducing some icons for the different checked states.
So I refactored it to just use those along with conditionals.
There are some holes in this, mainly around the indeterminate state of the checkbox (it'll require some JS), but we wanted to just implement it and move on until we need that state for some reason.
Along with these changes we also got rid of the need for 'CheckboxInput' component (checkbox + text).
We'll revisit that when we get to a form element.

PR #10, opened 2025-03-03
Updated some icons and added alphabetical sorting to sprite generation script

shaunrfox:


Etwigg (approved):
How many icons does one man need?

PR #23, opened 2025-03-11
Icon component refactor

### LINKED ISSUE #12: Icon
[Icon Figma](https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=62-2&m=dev&t=jMdbiIFKXjnDclFE-1)
>[!IMPORTANT]
>Important bit for this component is using the size prop in the viewbox!

Yash-CETEC:
Issue: https://github.com/Cetec-ERP/Cetec-Design-System/issues/12

### DIFF
--- a/src/components/Box/Box.tsx
+++ b/src/components/Box/Box.tsx
@@ -1,8 +1,8 @@
-import React, { 
+import React, {
+  ComponentPropsWithoutRef,
   type ElementType,
-  type AllHTMLAttributes,
   createElement,
-  } from 'react';
+} from 'react';
 import { box, type BoxVariantProps } from '@styled-system/recipes';
 import type { SystemStyleObject } from '@styled-system/types';
 import { cx } from '@styled-system/css';
@@ -10,22 +10,21 @@ import { splitProps } from '~/utils/splitProps';
 /*
  * Imports from recipes are placeholders for if we want to add some kind of styling to Box
  */
-export type BoxProps = Omit<AllHTMLAttributes<HTMLElement>, 'as'> & SystemStyleObject & BoxVariantProps & {
+export type BoxProps = Omit<ComponentPropsWithoutRef<ElementType>, 'as'> & SystemStyleObject & BoxVariantProps & {
   as?: ElementType;
 }
 
 export const Box: React.FC<BoxProps> = (
-  { 
-    as = "div", 
+  {
+    as = "div",
     ...props
   }
 ) => {
-  as = typeof as === "string" && as.length > 0 ? as : 'div';
-  const [ className, otherProps ] = splitProps(props);
+  const [className, otherProps] = splitProps(props);
   const comboClassName = cx(box({}), className);
   return (
     createElement(as, {
-      className: {comboClassName},
+      className: { comboClassName },
       ...otherProps
     })
   );
--- a/src/components/Box/Box.tsx
+++ b/src/components/Box/Box.tsx
@@ -22,7 +22,7 @@ export const Box: React.FC<BoxProps> = (
 ) => {
   as = typeof as === "string" && as.length > 0 ? as : 'div';
   const [ className, otherProps ] = splitProps(props);
-  const comboClassName = cx(box({}), className as string);
+  const comboClassName = cx(box({}), className);
   return (
     createElement(as, {
       className: {comboClassName},
--- a/src/components/Code/Pre.tsx
+++ b/src/components/Code/Pre.tsx
@@ -23,7 +23,7 @@ export const Pre: React.FC<PreProps> = (
       as="pre"
       className={cx(
         pre({}),
-        className as string,
+        className,
       )}
       {...otherProps}
     >
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -30,9 +30,9 @@ export const Icon: React.FC<IconProps> = ({
     <Box
       as={'svg'}
       name={name}
-      viewBox='0 0 24 24'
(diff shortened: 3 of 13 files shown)


Yash-CETEC:
@Etwigg and @shaunrfox please take a look at this and share your thoughts on it

src/components/Box/Box.tsx
```diff
@@ -1,31 +1,30 @@
-import React, { 
+import React, {
+  ComponentPropsWithoutRef,
   type ElementType,
-  type AllHTMLAttributes,
```
Etwigg, line 3: Why did you make this change?

src/components/Code/Pre.tsx
```diff
@@ -23,7 +23,7 @@ export const Pre: React.FC<PreProps> = (
       as="pre"
       className={cx(
         pre({}),
-        className as string,
```
Etwigg, line 26: I was getting TS errors without 'as string'

src/components/Icon/Icon.tsx
```diff
@@ -16,26 +19,24 @@ export type AllowedIconSizes = keyof typeof numericSizes;
 export type IconProps = Omit<BoxProps, 'size'> & SVGAttributes<SVGElement> & {
   name: IconNamesList;
   size?: AllowedIconSizes;
-  fill?: string;
+  fill?: ConditionalValue<WithEscapeHatch<ColorToken | `var(--${string})`>>;
```
Etwigg: ColorToken is probably enough here, right?

src/components/Icon/Icon.tsx
```diff
@@ -16,26 +19,24 @@ export type AllowedIconSizes = keyof typeof numericSizes;
...
   return (
     <Box
-      as={'svg'}
+      as="svg"
       name={name}
-      viewBox='0 0 24 24'
+      viewBox={`0 0 ${size} ${size}`}
       xmlns="http://www.w3.org/2000/svg"
-      className={cx(icon({ size: size as AllowedIconSizes, fill }), className)}
+      className={cx(icon({ size: size, fill }), className)}
```
Etwigg, line 40:
```suggestion
      className={cx(
        icon({ size: size, fill }), 
        className
      )}
```

src/utils/splitProps.ts
```diff
@@ -1,7 +1,9 @@
 import { splitCssProps } from '@styled-system/jsx';
 import { css } from '@styled-system/css';
 
-export function splitProps (props: object) {
+export const splitProps = (
+  props: Record<string, any>
+): [string, Record<string, any>] => {
```
Etwigg, line 6: Is this necessary?

Etwigg (changes requested):

src/components/Code/Pre.tsx
Yash-CETEC: I have declared className as string in the splitProps function return value hence we will not be required to explicitly set as className every time now

src/components/Box/Box.tsx
Yash-CETEC: AllHTMLAttributes does not include svg attributes so to include svg attributes as well I have replaced `AllHTMLAttributes` with `ComponentPropsWithoutRef`

Etwigg (dismissed):
Not seeing any errors, looks good.

src/utils/splitProps.ts
Yash-CETEC: I am assigning return type here so we don't need to declare return value type at every instance like className will be considered string every place now

src/components/Icon/Icon.tsx
```diff
@@ -16,26 +19,24 @@ export type AllowedIconSizes = keyof typeof numericSizes;
...
-  const className = css(cssProp, styleProps);
+  const [className, otherProps] = splitProps(props);
 
   return (
     <Box
-      as={'svg'}
+      as="svg"
       name={name}
-      viewBox='0 0 24 24'
+      viewBox={`0 0 ${size} ${size}`}
```
shaunrfox: This needs to remain `viewBox='0 0 24 24'`. The viewBox is the dimensions that the vectors are based on. All of the icons are built in a 24px square, so this should never change.

shaunrfox (changes requested):
Besides my one comment, this looks good.

src/components/Icon/Icon.tsx
Etwigg: I thought we had talked about using size for this and defaulting it to 24?

Yash-CETEC: @shaunrfox and @Etwigg  size prop will allow every value of `AllowedIconSizes` are we sure that by keeping `viewBox='0 0 24 24'` it will not have any issue in scaling when we use any value from `AllowedIconSizes` 
- if yes and we are sure that we are not going to extend the `AllowedIconSizes` values in future in such way in which will not have any issue with scaling of 24x24 then I am fine with keeping viewBox value static `'0 0 24 24'`
- else we need to think over it how viewBox will be changed based on size.

shaunrfox: The viewbox for the icons should never change, and will respond appropriately to the width/height/size values we pass in. All the icons are designed on a 24px grid and if we need any SVGs that don't align with that, they will be handled another way.

Yash-CETEC: Okay reverting the changes.

Yash-CETEC:
@shaunrfox can you take a look at this and share your thoughts on it?

shaunrfox (approved):
Looks good to me!

Etwigg (approved):

PR #25, opened 2025-03-26
Button component refactor

### LINKED ISSUE #13: Button
[Button Figma](https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=643-2808&m=dev&t=jMdbiIFKXjnDclFE-1)

>[!IMPORTANT]
>We want to use conditionals from panda as much as possible here. The Figma contains both normal Button and IconButton, but they're two different components in the code!

Yash-CETEC:
Issue: https://github.com/Cetec-ERP/Cetec-Design-System/issues/13


src/components/Button/Button.tsx
```diff
@@ -1,52 +1,57 @@
...
-export type ButtonProps = BoxProps & ButtonVariantProps & {
-  variant?: 'primary' | 'standard' | 'hollow' | 'ghost' | 'cta' | 'danger';
-  href?: string;
-  loading?: boolean;
-  className?: string;
-  children?: string | React.ReactNode;
-  disabled?: boolean;
-};
+type buttonVariants =
+  | 'primary'
```
Etwigg, line 9:
```suggestion
  'primary'
```

Etwigg: I think these can be on one line.

Etwigg (dismissed):
This all looks fine to me. I was expecting css changes with this as well though. Are you planning that in a later PR?

Etwigg:
@shaunrfox

src/components/Button/Button.tsx
Yash-CETEC: This is an ESLint suggestion which is automatically implemented by it.

Yash-CETEC:
> This all looks fine to me. I was expecting css changes with this as well though. Are you planning that in a later PR?

The PR was raised after Icon Component Related PR was raised but code was done 3 weeks ago at that time figma for button was not available to me so let me take a look at figma changes and get back to you.

Etwigg (changes requested):
I think some of the css is still off? Maybe the App file just isn't using the correct thing? 
For example I'm seeing the primary button as red here:
image

Yash-CETEC:
> I think some of the css is still off? Maybe the App file just isn't using the correct thing? For example I'm seeing the primary button as red here: image

Screenshot 2025-04-01 at 12 52 50 PM
@Etwigg, here the CSS property to change the background colour to red was already added, so I kept it assuming it was put there as an example to show how CSS is added. If there is no reason to keep it let me know I will remove it as well.
Or I can also create one more extra button to show that use case on another button.

src/styles/tokens.ts
```diff
@@ -370,22 +353,22 @@ export const durations = {
...
-  wide: {
+  '0.025%': {
     value: '0.025em',
   },
-  wider: {
+  '0.05%': {
     value: '0.05em',
   },
-  widest: {
+  '0.1%': {
```
Etwigg: not a fan of this change

Etwigg: @shaunrfox What do we want to do with letter spacing tokens?

Etwigg:
> > I think some of the css is still off? Maybe the App file just isn't using the correct thing? For example I'm seeing the primary button as red here: image
> 
> Screenshot 2025-04-01 at 12 52 50 PM @Etwigg, here the CSS property to change the background colour to red was already added, so I kept it assuming it was put there as an example to show how CSS is added. If there is no reason to keep it let me know I will remove it as well. Or I can also create one more extra button to show that use case on another button.

Yeah it's probably a good idea to just add a button for each variant on here.

Yash-CETEC:
> > > I think some of the css is still off? Maybe the App file just isn't using the correct thing? For example I'm seeing the primary button as red here: image
> > 
> > 
> > Screenshot 2025-04-01 at 12 52 50 PM @Etwigg, here the CSS property to change the background colour to red was already added, so I kept it assuming it was put there as an example to show how CSS is added. If there is no reason to keep it let me know I will remove it as well. Or I can also create one more extra button to show that use case on another button.
> 
> Yeah it's probably a good idea to just add a button for each variant on here.

@Etwigg changes is done.

src/styles/tokens.ts
Yash-CETEC: @Etwigg and @shaunrfox either we can make changes here or we can change the letter-spacing values in Figma.

src/components/Button/Button.tsx
```diff
@@ -1,52 +1,57 @@
...
+    href?: string;
+    loading?: boolean;
+    className?: string;
+    children?: string | ReactNode;
+    disabled?: boolean;
+    type?: buttonTypes;
   };
 
+export const Button: FC<ButtonProps> = ({
+  variant = 'primary',
```
shaunrfox: This needs to be changed to default to 'standard'. This aligns with the defaults set in the recipe. (I looked at the git history and think this was my mistake)

src/components/Button/Button.tsx
```diff
@@ -1,52 +1,57 @@
...
+  size,
+  href,
+  children,
+  loading = false,
+  disabled,
```
shaunrfox, line 35: If we're adding loading=false, should we also have disabled=false?

src/components/Button/ButtonContent.tsx
```diff
@@ -1,31 +1,32 @@
...
-        top="0"
-        left="0"
-        right="0"
-        bottom="0"
-        placeItems="center"
-      >
-        <Spinner />
-      </Grid>
-    )}
+    <Box
```
shaunrfox: Why this change? I'd prefer to only render this if the state is true, rather than always rendering it, but hiding it.

src/styles/tokens.ts
```diff
@@ -250,24 +250,7 @@ export const fontSizes = {
...
-  default: {
-    value: '1em + 0.5rem',
-  },
-  loose: {
-    value: '1em + 0.75rem',
-  },
-  looser: {
-    value: '1em + 0.875rem',
-  },
+  ...numericSizes
```
shaunrfox: These need to remain. This is the way I want the line-heights to work, but Figma doesn't support this kind of definition, so that's why the numeric scale is used in Figma.

src/styles/tokens.ts
shaunrfox: These should remain as they were.

Yash, I'm not sure where you're seeing any letter-spacing values in Figma. I didn't think I'd defined any.

src/recipes/button.ts
```diff
@@ -3,30 +3,31 @@ import { defineRecipe } from '@pandacss/dev';
...
   alignItems: 'center',
-  gap: 4,
+  gap: '4',
   fontFamily: 'sans',
-  fontSize: 16,
+  fontSize: '16',
   fontWeight: 'medium',
-  lineHeight: 'default',
-  borderWidth: 1,
+  lineHeight: '24',
```
shaunrfox: This needs to remain 'default'

src/recipes/button.ts
```diff
@@ -3,30 +3,31 @@ import { defineRecipe } from '@pandacss/dev';
...
+  borderRadius: '4',
+  outlineWidth: '2',
   outlineStyle: 'solid',
   outlineColor: 'transparent',
-  outlineOffset: 1,
+  outlineOffset: '1',
   textDecoration: 'none',
   whiteSpace: 'nowrap',
   cursor: 'pointer',
+  letterSpacing: '0%',
```
shaunrfox: This shouldn't be needed

src/recipes/button.ts
```diff
@@ -102,13 +103,14 @@ const buttonVariants = {
...
         bg: { base: 'slate.90', _dark: 'slate.5' },
-        color: { base: 'slate.5', _dark: 'slate.90' },
+        color: { base: 'slate.0', _dark: 'slate.90' },
         borderColor: 'transparent',
       },
     },
     ghost: {
       bg: 'transparent',
-      color: { base: 'slate.90', _dark: 'slate.5' },
+      borderColor: 'transparent',
```
shaunrfox: This is already set in the buttonBase styles above, so we shouldn't need to repeat it here.

src/components/Button/Button.tsx
```diff
@@ -1,52 +1,57 @@
...
-        type={asComponent === 'button' ? 'button' : undefined}
-        {...props}
-      >
-        <>
-          <ButtonContent loading={!!loading}>{children}</ButtonContent>
-        </>
-      </Box>
-    );
+export type ButtonProps = BoxProps &
+  ButtonVariantProps & {
```
shaunrfox, line 19: It looks like we still need to handle the 'selected' state – see the button recipe

src/recipes/button.ts
shaunrfox: Actually, this looks like it was moved from the _selected styles. This might need to be put back once the selected state is working

src/recipes/button.ts
```diff
@@ -169,26 +170,27 @@ export const buttonRecipe = defineRecipe({
   variants: {
     ...buttonVariants,
     size: {
-      standard: {
+      medium: {
         fontSize: '16',
-        py: '3',
+        py: '4',
```
shaunrfox: The padding here should remain – same with lines 180-181. This is a gotcha with Figma where the border is not necessarily included in the box calculations. These values are what they were to account for the transparent 1px border.

src/recipes/button.ts
```diff
@@ -169,26 +170,27 @@ export const buttonRecipe = defineRecipe({
...
-        py: '7',
-        px: '14',
+        py: '8',
+        px: '16',
       },
       small: {
         fontSize: '14',
         py: '1',
         px: '8',
+        lineHeight: '22',
```
shaunrfox: This should be able to inherit 'default' from line 18 above. Since the values are based on the font sizes, this already calculates to 22.

shaunrfox (changes requested):

src/styles/tokens.ts
Etwigg: Are the ones in the tokens file the panda defaults? @shaunrfox 
I think we only use the default 'normal' (0em) in all of the existing figma designs.

src/recipes/button.ts
Etwigg: That's kinda tricky. Can we do anything to make that more clear in the future? Or do you think it's uncommon enough to not be a problem?

src/components/Button/Button.tsx
Yash-CETEC: Well, that was done for some past you we can remove it now else we can let it be there. It won't make much difference

src/components/Button/ButtonContent.tsx
Yash-CETEC: For using both the difference is that rather than based on the state introducing a new component or removing new component from actual DOM, I am trying just to do the CSS changes for the same

src/styles/tokens.ts
Yash-CETEC: @Etwigg and @shaunrfox This need to be adequately addressed, we expect both of the values mentioned in Figma and in the design system should be the same.

src/styles/tokens.ts
Yash-CETEC: @Etwigg and @shaunrfox I can see the letter spacing value in Figma that's why it is implemented.

Screenshot 2025-04-04 at 6 22 33 PM

src/components/Button/Button.tsx

src/recipes/button.ts
Yash-CETEC: I can see the letter-spacing value in the Figma so I implemented them. You can make text focus in Figma by double-clicking to see the value.


Screenshot 2025-04-04 at 6 36 10 PM

src/recipes/button.ts
Yash-CETEC: Done!

src/recipes/button.ts
Yash-CETEC: Done! Removed redundant same properties from all the variants except `hollow` variant

src/components/Button/Button.tsx
Yash-CETEC: Hey @shaunrfox, can you elaborate on this? As far as I can see only the `cta` and `danger` variants doesn't have `_selected` property on them.

src/styles/tokens.ts
Yash-CETEC: @Etwigg and @shaunrfox reverted the changes for letterSpacing value

src/components/Button/Button.tsx
Yash-CETEC: @shaunrfox removed changes from loading Prop as well.

src/styles/tokens.ts
Yash-CETEC: @Etwigg and @shaunrfox , I have reverted back the changes.

src/recipes/button.ts
Yash-CETEC: @shaunrfox , I have reverted back the value to `default`

src/components/Button/Button.tsx
Yash-CETEC: @shaunrfox , I think we don't need to define anything here as in the button recipe we have already defined

```
defaultVariants: {
    variant: 'standard',
    size: 'medium',
  }
```
so I don't think we need to mention them here anymore. I have already tested by removing the default value for the variant here in the component and it is working as per expectations so I am removing it. don't need to mention default value at multiple places.

src/recipes/button.ts
Yash-CETEC: Done! removed...

src/components/Button/Button.tsx
```diff
@@ -27,7 +27,7 @@ export type ButtonProps = BoxProps &
   };
 
 export const Button: FC<ButtonProps> = ({
-  variant = 'standard',
```
Etwigg: Why remove default variant?

src/components/Button/Button.tsx
Etwigg: Looks correct to me, the recipe has the _selected variants mentioned in figma.

src/recipes/button.ts
```diff
@@ -169,15 +161,15 @@ export const buttonRecipe = defineRecipe({
...
         fontSize: '16',
         py: '3',
         px: '12',
       },
       large: {
         fontSize: '16',
-        py: '7',
-        px: '14',
+        py: '8',
+        px: '16',
```
Etwigg: Does this padding need to remain the same as well?

src/components/Button/ButtonContent.tsx
Etwigg: What's the actual performance difference here?

Etwigg (changes requested):

src/components/Button/Button.tsx
Yash-CETEC: We have already included the default variant directly in the recipe, so it doesn’t need to be added again here.

Screenshot 2025-04-10 at 1 26 08 PM

src/components/Button/ButtonContent.tsx
Yash-CETEC: My bad!! After going through the details again I do find the original implementation better. So reverting back the changes.

src/recipes/button.ts
Yash-CETEC: @shaunrfox can you help me on this?

src/recipes/button.ts
shaunrfox: Yeah, I'll look into this some more. It shouldn't be something we run into very often, except for these foundational elements that we have specific height targets for.

src/recipes/button.ts
shaunrfox: Ok, I apologize for the confusion. Here are the final padding values for these button styles, which should now be properly reflected in Figma as well.

**in buttonRecipe:**
```
size: {
  medium: {
    fontSize: '16',
    py: '3',
    px: '10',
  },
  large: {
    fontSize: '16',
    py: '7',
    px: '12',
  },
  small: {
    fontSize: '14',
    py: '0',
    px: '8',
  }
}
```

**in iconButtonRecipe:**
```
size: {
  medium: {
    fontSize: '16',
    p: '3',
  },
  large: {
    fontSize: '16',
    p: '7',
  },
  small: {
    fontSize: '14',
    p: '0',
  },
}
```

Yash-CETEC: Done!

Etwigg (approved):

PR #27, opened 2025-04-16
Icon button component refactor

### LINKED ISSUE #15: IconButton
[IconButton Figma](https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=643-2808&m=dev&t=zcVxkrV3CMOJ1qSw-1)
>[!IMPORTANT]
>IconButton shares a figma file with normal Button, but they're separate components in the code

Yash-CETEC:
Issue: https://github.com/Cetec-ERP/Cetec-Design-System/issues/15


src/components/IconButton/IconButton.tsx
```diff
@@ -7,34 +7,53 @@ import {
...
- * Define the polymorphic component type for IconButton.
- */
-type IconButtonComponent = <E extends React.ElementType = 'button'>(
-  props: IconButtonProps<E> & { ref?: React.ForwardedRef<Element> },
-) => JSX.Element;
+type IconButtonComponent = {
+  buttonText?: string;
+  iconName: IconNamesList;
+  iconPosition: IconPositions;
+  size: AllowedIconSizes;
```
Etwigg: Why use the additional type here? We can make the component from the IconButtonProps type, right?

src/components/IconButton/IconButton.tsx
```diff
@@ -7,34 +7,53 @@ import {
...
+      gap="8"
+      flexDirection={iconPosition === 'right' ? 'row' : 'row-reverse'}
+    >
+      <Box as="span" {...(!buttonText && { display: 'none' })}>
+        {buttonText}
+      </Box>
+      <Icon name={iconName} size={size} />
+    </HStack>
+  );
+};
```
Etwigg: Is this supposed to be like the normal Button's ButtonContent component? If so, it should follow that pattern and be a separate component.

src/components/IconButton/IconButton.tsx
```diff
@@ -44,47 +63,71 @@ type IconButtonComponent = <E extends React.ElementType = 'button'>(
...
+  children,
+  loading,
+  disabled,
+  iconName,
+  type,
+  buttonText,
+  iconPosition = 'right',
+  ...props
+}: IconButtonProps) => {
+  const trulyDisabled = loading || disabled;
```
Etwigg: can we call this isDisabled or something?

src/components/IconButton/IconButton.tsx
```diff
@@ -44,47 +63,71 @@ type IconButtonComponent = <E extends React.ElementType = 'button'>(
...
+
+    return isThereIconInChild;
+  }, [isChildrenExists, children]);
+
+  if (
+    (isChildrenExists && !containsIcon) ||
+    (!isChildrenExists && iconName === undefined)
+  ) {
+    console.error(
+      'Please provide IconButton with at least one Icon component in children or proper iconName',
```
Etwigg: Why don't we just make icon name a required prop instead of doing all of this usememo stuff?

src/components/IconButton/IconButton.tsx
```diff
@@ -44,47 +63,71 @@ type IconButtonComponent = <E extends React.ElementType = 'button'>(
...
+          ) : (
+            <IconButtonComponent
+              buttonText={buttonText}
+              iconName={iconName!}
+              iconPosition={iconPosition}
+              size={size === 'small' ? '22' : '24'}
+            />
+          )}
+        </ButtonContent>
+      </ButtonContent>
```
Etwigg: Why are we nesting this?
And using button content here in combination with your new icon button component (which seems like the icon button version of button content) seems odd?

Etwigg (changes requested):

src/components/IconButton/IconButton.tsx
Yash-CETEC: Done!

src/components/IconButton/IconButton.tsx
Yash-CETEC: I have created this new IconButtonComponent to accommodate the standard design pattern mentioned in Figma, where only the icon is there, or the icon is at the right or left of the text.

src/components/IconButton/IconButton.tsx
Yash-CETEC: I have created this type not for the `IconButton` component but for the `IconButtonComponent`

src/components/IconButton/IconButton.tsx
Yash-CETEC: The previous design allowed the user to pass a custom children component directly instead of the current Figma, which has a fixed structure. To enable that custom structure as well, I have provided here a condition such that if the Developer provided a child component directly, this code will check whether there is an Icon part of it. 

- @Etwigg and @shaunrfox, do we only want to allow the following structure or allow a custom child component as well?

Screenshot 2025-04-21 at 3 29 22 PM

src/components/IconButton/IconButton.tsx
Yash-CETEC: In this implementation, I utilise the pre-existing ButtonContent component, enabling the conditional rendering of either a custom child component or an IconButtonComponent, depending on the presence of child elements.

src/components/IconButton/IconButton.tsx
Etwigg: Don't worry about custom child component

Etwigg:
@Yash-CETEC 
@shaunrfox Has said that the IconButton is supposed to be only for when the button is nothing but a single icon.
Please make adjustments accordingly, and apologies for the confusion.
Thanks

src/components/Button/Button.tsx
```diff
@@ -30,6 +20,7 @@ export const Button: FC<ButtonProps> = ({
   variant,
   size,
   href,
+  className: customeClassName = '',
```
Etwigg:
```suggestion
  className: customClassName = '',
```

src/components/Button/Button.tsx
```diff
@@ -45,7 +36,7 @@ export const Button: FC<ButtonProps> = ({
       as={href ? 'a' : 'button'}
       disabled={trulyDisabled}
       aria-disabled={trulyDisabled}
-      className={cx(button({ variant, size }), className)}
+      className={cx(button({ variant, size }), className, customeClassName)}
```
Etwigg:
```suggestion
      className={cx(button({ variant, size }), className, customClassName)}
```

src/components/IconButton/IconButton.tsx
```diff
@@ -1,90 +1,83 @@
...
-      iconName?: IconNamesList;
-    };
-
-/**
- * Define the polymorphic component type for IconButton.
- */
-type IconButtonComponent = <E extends React.ElementType = 'button'>(
-  props: IconButtonProps<E> & { ref?: React.ForwardedRef<Element> },
-) => JSX.Element;
+import { Icon } from '~/components/Icon';
```
Etwigg, line 9: No type?

src/components/IconButton/IconButton.tsx
```diff
@@ -1,90 +1,83 @@
...
+  };
 
-    // Correctly infer the size for the Icon based on the size prop
-    const iconSize: SizeToken = size === 'small' ? '22' : '24';
+const isInvalidIconButton = (children: ReactNode): boolean => {
+  if (children == null) return true;
+  if (Children.count(children) !== 1) return true;
+  const onlyChild = Children.only(children);
+  return !(isValidElement(onlyChild) && onlyChild.type === Icon);
+};
```
Etwigg: Why do this? Aren't we handling this with props?

src/components/IconButton/IconButton.tsx
```diff
@@ -1,90 +1,83 @@
...
+      'IconButton: expected a single <Icon> child but received invalid children.',
     );
-  },
-) as IconButtonComponent;
+    return null;
+  }
+
+  return (
+    <Box
+      as={href ? 'a' : 'button'}
```
Etwigg, line 44: Where does this come from? Why handle this as an anchor?

src/components/IconButton/IconButton.tsx
```diff
@@ -1,90 +1,83 @@
...
+      disabled={isDisabled}
+      aria-disabled={isDisabled}
+      className={cx(iconButton({ variant, size }), className, customClassName)}
```
Etwigg: I asked about customClassName in the other PR, just want to note it here too.

Etwigg (changes requested):
Some questions!

src/components/Button/Button.tsx
Yash-CETEC: Done in another PR regarding the button component.

src/components/Button/Button.tsx
Yash-CETEC: Done in another PR regarding the button component.

src/components/IconButton/IconButton.tsx
Yash-CETEC: We do that, but TypeScript can only display and does not prevent the user from adding multiple elements or even non-icon elements to the IconButton component. Therefore, this code will verify the children and their types during runtime and will not show that component on the screen.

Is it fine for me to put this type of restriction on this kind of very specific, structured and use case component? @Etwigg and @shaunrfox, please share your thoughts on it.

Screenshot 2025-04-28 at 5 18 39 PM

src/components/IconButton/IconButton.tsx
Yash-CETEC: As per Figma documentation https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1263-1176&m=dev, IconButton and Button components are considered the same, with all props being are same for both components. Hence, I am assuming that href can also be passed to the IconButton as well and added needful changes in it as well.

src/components/IconButton/IconButton.tsx
Yash-CETEC: Given answer there but here is the link to that answer: https://github.com/Cetec-ERP/Cetec-Design-System/pull/28#discussion_r2063406491

src/components/IconButton/IconButton.tsx
Etwigg: I don't feel it's necessary (at least not right now?)

src/components/Button/Button.tsx
```diff
@@ -4,20 +4,10 @@ import { Box, type BoxProps } from '~/components/Box';
...
-type buttonVariants =
-  | 'primary'
-  | 'standard'
-  | 'hollow'
-  | 'ghost'
-  | 'cta'
-  | 'danger';
-
-type buttonTypes = 'submit' | 'reset' | 'button';
+import { buttonTypes } from '~/recipes/button';
```
shaunrfox: As I mentioned in the Button PR, I'd prefer to keep the type definitions with the component code.

src/components/IconButton/IconButton.tsx
shaunrfox: This is a hard line to define, but I'd prefer to avoid being restrictive in this way unless it becomes a problem. I expect Typescript linting and code reviews to cover cases like this.

shaunrfox (changes requested):

src/components/IconButton/IconButton.tsx
Yash-CETEC: Done!

src/components/Button/Button.tsx

src/components/IconButton/IconButton.tsx
```diff
@@ -1,90 +1,60 @@
...
+  };
 
-    // If no children are provided and an iconName is specified, render the Icon automatically.
-    const content =
-      children ?? (iconName ? <Icon name={iconName} size={iconSize} /> : null);
+export const IconButton: FC<IconButtonProps> = ({
+  variant,
+  size,
+  href,
+  className: customClassName = '',
```
shaunrfox: This can be removed, per our discussion

shaunrfox (changes requested):

src/components/IconButton/IconButton.tsx
Yash-CETEC: Done!

Etwigg (approved):

PR #28, opened 2025-04-24
Button component changes

### LINKED ISSUE #13: Button

Yash-CETEC:
Issue: https://github.com/Cetec-ERP/Cetec-Design-System/pull/25


src/components/Button/Button.tsx
```diff
@@ -4,20 +4,10 @@ import { Box, type BoxProps } from '~/components/Box';
...
-  | 'ghost'
-  | 'cta'
-  | 'danger';
-
-type buttonTypes = 'submit' | 'reset' | 'button';
+import { buttonTypes } from '~/recipes/button';
 
 export type ButtonProps = BoxProps &
   ButtonVariantProps & {
-    variant?: buttonVariants;
```
Etwigg, line 20: Why remove this?

src/components/Button/Button.tsx
```diff
@@ -45,7 +36,7 @@ export const Button: FC<ButtonProps> = ({
       as={href ? 'a' : 'button'}
       disabled={trulyDisabled}
       aria-disabled={trulyDisabled}
-      className={cx(button({ variant, size }), className)}
+      className={cx(button({ variant, size }), className, customeClassName)}
```
Etwigg: Can you explain customeClassName?
Also, I think it should be custom instead of custome.

Etwigg (changes requested):
A few questions

Etwigg:
Side note: Button have their own text styles outside of Heading/Body/Mono.
It's just a slightly heavier weight, so we need to make sure it's included in the recipe.

src/components/Button/Button.tsx
Yash-CETEC: `ButtonVariantProps` already contains `variant` and `size` props inside it, which are not required to be re-created again.

src/components/Button/Button.tsx
Yash-CETEC: Changed name to `customClassName`.

As for the requirement, as you can see here in Figma https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1263-1108&m=dev, we are allowing developers to pass `className` as a prop. So to differentiate between the className made using the `splitProps` function and the `className` prop, I have assigned the name `customClassName` to the className prop.

Screenshot 2025-04-28 at 4 17 08 PM

Yash-CETEC:
> Side note: Button have their own text styles outside of Heading/Body/Mono. It's just a slightly heavier weight, so we need to make sure it's included in the recipe.

@Etwigg, according to the Figma specifications, the button text should utilise a font weight of 500. In our design system, this corresponds to the medium token value, which is already incorporated within the button component’s recipe.

src/components/Button/Button.tsx
Etwigg: I don't think we need className as a prop here, right? We get that by default without having to specify it in the props. Using customClassName, etc. seems like an over complication?

src/components/Button/Button.tsx
Etwigg: Don't we want the exclude here also? Exclude ButtonVariantProps from BoxProps?

src/recipes/button.ts
```diff
@@ -1,5 +1,7 @@
 import { defineRecipe } from '@pandacss/dev';
 
+export type buttonTypes = 'submit' | 'reset' | 'button';
```
shaunrfox: I think we should be keeping the recipes specific to their intended use – styles and style props. Moving these types out of the component to here feels like a non-obvious place to look. It's also atypical to reference anything in the `~/recipes/` directory since it's intended to be compiled into the `@styled-system/` folders.

src/components/Button/Button.tsx
```diff
@@ -4,20 +4,10 @@ import { Box, type BoxProps } from '~/components/Box';
...
-type buttonVariants =
-  | 'primary'
-  | 'standard'
-  | 'hollow'
-  | 'ghost'
-  | 'cta'
-  | 'danger';
-
-type buttonTypes = 'submit' | 'reset' | 'button';
+import { buttonTypes } from '~/recipes/button';
```
shaunrfox: As noted in the recipe, I'd prefer that we keep the type definitions in this file.

src/recipes/button.ts
```diff
@@ -175,6 +177,10 @@ export const buttonRecipe = defineRecipe({
         fontSize: '14',
         py: '0',
         px: '8',
+        '& svg': {
+          mt: '-1',
+          mb: '-1'
+        }
```
shaunrfox, line 181: Thanks for adding this. The only thing I see that's missing from this PR is a small change to the ButtonContent component, [like in my test branch](https://github.com/Cetec-ERP/Cetec-Design-System/compare/main...sf-button-changes#diff-05bf8cf8690ab774b4cdec22b2eb0967fad978b7ce39efd7b634e8544926791eR16).

shaunrfox (changes requested):

src/components/Button/Button.tsx
shaunrfox: Yeah, I agree. Since we're spreading props, this should already be inherently supported, right?

src/components/Button/Button.tsx
Yash-CETEC: `&` create a composite type of BoxProps and ButtonVariant that allows the merging of similar properties into one property, which can be assigned all the possible values declared in both types. I removed the variant property because it was explicitly created by again by us in the last object, which has the same allowed value as `ButtonVariantProps` variant property, so it was not providing any additional benefit and was creating redundancy.

src/components/Button/Button.tsx
Yash-CETEC:
```
@Etwigg and @shaunrfox, Following scenarios are needed to be considered:
1. Are we going to allow developers pass `className` as prop to the Button Component or not? 
	- If no, then there is no requirement of className as prop and className as type in ButtonProps?
	- If yes, following scenarios to considered
		1. The className provided as prop is going to completely override the style provided by the recipe?
			- If yes, than the point mentioned by @shaunfox in above comment is applicable and we don't need to explicitly extract className as prop and will override the className provided by the splitProp function.
			- If no, meaning the className provided by developer should not completely override the existing style but add extra property in that case, there are two ways to do that:
				1. If we need to do some changes for the rare instance for limited amount of property we can directly pass the style property as prop and `splitProps` function which internally uses `splitCssProps` function to merge them in existing className created by splitProps function
				2. If the property changes we will be implemented repeatedly for considerable times at that time then we can handle them in recipe
P.S. Currently I am not able consider the scenarios where className is required to be passed as prop. Can you guys help me understand the scenarios where it may be required? Maybe I can come up with alternate approach so that we don't need to do this?
```

src/recipes/button.ts
Yash-CETEC: Done!

src/components/Button/Button.tsx

src/recipes/button.ts
Yash-CETEC: Done the needful changes!

src/components/Button/Button.tsx
Etwigg: We can just remove it from props.

shaunrfox: I agree that we should just remove it from the props. It's not something that we intend to use, given all of the styling options we've made available through Panda. I don't think we need to go to any effort to restrict it, though. Removing it from the props seems like enough.

src/components/Button/Button.tsx
```diff
@@ -5,35 +5,25 @@ import { button, type ButtonVariantProps } from '@styled-system/recipes';
...
     disabled?: boolean;
-    type?: buttonTypes;
+    type?: 'submit' | 'reset' | 'button';
   };
 
 export const Button: FC<ButtonProps> = ({
   variant,
   size,
   href,
+  className: customClassName = '',
```
shaunrfox: This can be removed, per our discussion

shaunrfox (changes requested):

src/components/Button/Button.tsx

Yash-CETEC:
@Etwigg and @shaunrfox all the needful changes mentioned in the pull request are done. Kindly review it and share your thoughts on it!

Etwigg (approved):

PR #42, opened 2025-05-12
Text component refactor

### LINKED ISSUE #17: Text
[Text Figma](https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=133-336&m=dev&t=HVgjRjczWEhQyxEg-1)

CSS refactor to align with the figma.

Yash-CETEC:
Issue: https://github.com/Cetec-ERP/Cetec-Design-System/issues/17


src/components/Text/Text.tsx
```diff
@@ -1,46 +1,31 @@
-import React, {
-  type ElementType,
-} from 'react';
+import React, { ReactNode, type ElementType } from 'react';
 import { Box, type BoxProps } from '~/components/Box';
-import type { FontToken, FontSizeToken, FontWeightToken } from '@styled-system/tokens';
```
Etwigg, line 5: Why are we getting rid of the token import and specifying those in the props?

src/components/Text/Text.tsx
```diff
@@ -1,46 +1,31 @@
...
+  as = 'p',
+  family,
+  italic,
+  bold,
+  underline,
+  size,
+  children,
+  ...props
+}: TextProps) => {
+  const [className, otherProps] = splitProps(props);
```
Etwigg: Why are we getting rid of weight?

src/recipes/text.ts
```diff
@@ -1,21 +1,25 @@
-import { defineRecipe } from '@pandacss/dev';
-import { fontSizes, fontWeights } from '~/styles/tokens';
+import { defineRecipe, SystemStyleObject } from '@pandacss/dev';
+import { OneOrMore } from '@styled-system/types';
```
Etwigg: What? We're getting rid of tokens and using some new type here?

src/recipes/text.ts
```diff
@@ -38,13 +42,64 @@ const textVariants = {
...
   },
   size: {
-    ...fontSizes,
-  },
-  weight: {
-    ...fontWeights,
+    lg: {},
+    md: {},
+    sm: {},
+    xs: {},
```
Etwigg: Very confused by this change. Getting rid of weight? Not using our font size tokens?

src/recipes/text.ts
```diff
@@ -38,13 +42,64 @@ const textVariants = {
...
 
+export type TextSizes = "lg" | "md" | "sm" | "xs"
+
+type TextCompoundVariants = {
+  family?: OneOrMore<"heading" | "mono" | "body">;
+  size?: OneOrMore<TextSizes>;
+  css: SystemStyleObject;
+};
+
+const textCompoundVariants: TextCompoundVariants[] = [
```
Etwigg: You want to replace the normal size prop with compound variants?
I think we're supposed to use textStyles for this instead. It's specifically for text (see header) and functions the same way afaik.

Etwigg (changes requested):

src/components/Text/Text.tsx
Yash-CETEC: TextVariantProps already includes `family` and `size` prop based on the pre-defined variantions of text provided in figma.
Based on family and size prop we have already pre-defined properties `font-family`, `font-size` and `font-weight`, so we don't need to pass `weight` anymore.
If, in an exceptional and rare case, we want to change the `font-family`, `font-size`, and `font-weight` properties, we can do so by directly passing them as props in the `Text` component, and the `splitProps` function will handle their implementation in the component.
@shaunrfox, I think the `weight` property also needs to be removed from the Figma description.

src/components/Text/Text.tsx
Yash-CETEC: As mentioned above, `family` for most of the cases of text component in figma weight is directly associated with the type of family provided and for any rare and exception case should we want to change we can directly pass `font-weight` as prop and `splitProps` function will take care of it's implementation

src/recipes/text.ts
Yash-CETEC: As mentioned in figma we have selected limited sizes and weights for our text component to maintain uniformity and consistancy, for any rare and exceptional case where we require to provide `font-size` and `font-weight` we can directly pass as prop to Text component and `splitProps` function will handle it. But I would like to emphasis that ideally we should not require to handle such rare and exceptional cases and if required it should be done seldomly.

src/recipes/text.ts
Yash-CETEC: Screenshot 2025-05-14 at 4 22 32 PM

Above is the code changes (where I have kept existing recipe changes for comparison) which will need to be implemented to accommodate all the existing props.
As you can see in above code I am able to handle family and size by existing textStyles `family.size` but I am not able to add parallel variant options like `bold`, `italic` and `underline` I have to add them conditional spread in text component this makes this logic non reusable.
Also, I have add `className` created by `splitProps` seperately to `className` prop.
I would recommend to follow `recipe` approach instead of `textStyles` which can allow us extend this style if require to add new variant option in future rather than doing new conditional addition of value.
In my opinion recipe approach is more clean, extendable and can hande different concern more efficienty if required in future.
P.S Recipe can do all the things that styles can do but wise versa is not true.
@shaunrfox, Can you also share your thoughts on this?

src/recipes/text.ts
Etwigg: But why make these empty objects?

src/recipes/text.ts
Etwigg: I think I didn't convey what I meant very well.
In /src/styles/tokens.ts there's a textStyles export. We use it in the heading recipe. I was saying we should do the same in the text recipe instead of the compound variants.

Yash-CETEC: I get your point @Etwigg , but I have already tried to implement existing `textStyles` in our component but as you can see that comes with it's limitations.

src/components/Text/Text.tsx
Yash-CETEC: reintroduced weights

src/recipes/text.ts
Yash-CETEC: reintroduced tokens back

src/recipes/text.ts
Yash-CETEC: removed empty objects and reintroduced fontSizes and fontWeights tokens

src/recipes/text.ts
Yash-CETEC: implemented based on our last discussion

src/recipes/text.ts
```diff
@@ -1,21 +1,59 @@
...
+type FontWeightKey = keyof typeof fontWeightTokens;
+const fontWeights = (Object.keys(fontWeightTokens) as FontWeightKey[]).reduce(
+  (accumulator, currentKey) => {
+    accumulator[currentKey] = {
+      fontWeight: fontWeightTokens[currentKey].value,
+    };
+    return accumulator;
+  },
+  {} as Record<FontWeightKey, Record<'fontWeight', number>>,
+);
```
Etwigg, line 33: Why are we doing this here?

src/styles/tokens.ts
```diff
@@ -493,25 +493,18 @@ const baseHeadingStyles = defineStyles({
...
 });
 
 const baseMonoStyles = defineStyles({
   value: {
     fontFamily: 'mono',
-    fontWeight: 'normal',
-    color: { base: 'slate.60', _dark: 'slate.20' },
-    lineHeight: 'default',
   },
 });
```
Etwigg, line 473: Are we using these base styles anywhere?

Yash-CETEC: These base styles are used in side `textStyles` only, but all three styles `baseHeadingStyles`, `baseBodyTextStyles` and `baseMonoStyles` uses values which are already defined in `textBase` so just removed redundant values.

src/recipes/text.ts
Yash-CETEC: Recipe can access token values inside properties but not the tokens directly. i.e we need assign `fontWeights` token value to `fontWeight` property inside a recipe. 
`...fontWeights` and `...fontSizes` will add values as it is in tokens file directly instead of assigning value to fontWeight

Recipe requires:

Screenshot 2025-05-16 at 12 20 17 PM


Where as spreading tokens directly inside recipe will do:

Screenshot 2025-05-16 at 12 12 25 PM

which will result in:

Screenshot 2025-05-16 at 12 14 53 PM

so above code takes all the key value from the `fontWeights` and `fontSizes` tokens and conver them into the format that recipe requires i.e 'tokenKey': { cssPropertyName: tokenValue }. so when size / weight applied from props following css will result in browser.

Screenshot 2025-05-16 at 12 18 44 PM

Etwigg: So, the sizes and weights never worked?

Etwigg: @shaunrfox

Yash-CETEC: > So, the sizes and weights never worked?

No for me when I tried to apply property, it was not working.

Etwigg: I thought we just needed value..hm

Etwigg (approved):
ok!
:shipit:

PR #43, opened 2025-05-12
fix: color tokens comprised of other tokens are now working properly

Yash-CETEC:


Etwigg (dismissed):

Etwigg (approved):
I'm not opposed to moving the semantic tokens out.

src/styles/semanticTokens.ts
```diff
@@ -0,0 +1,110 @@
+export const colors = {
```
shaunrfox, line 1: I'm fine with this change, but it also feels unnecessary for it to be a part of this PR, since the code was already functioning properly. Going forward I'd like us to consider two things:

1. Are there rules we should put in place for when we should or shouldn't abstract out code like this?
2. I want to limit the number of changes we're making to the core style-related files, like tokens and the panda config. Changes to these files can have broad-reaching implications, so we should be very careful with them going forward.

shaunrfox (approved):

Issue #4, opened 2025-02-07
Divider Component


atkinsdavid:
CodePen: ?
Figma Design: ?

@Etwigg might need to add something useful here.

Etwigg:
There's no codepen or figma for this one!
It's just a line that extends to the size of its container.
Needs to be vertical or horizontal and take color and weight (thickness of the line).

PR #44, opened 2025-05-12
Divider Component

atkinsdavid:
Currently WIP for https://github.com/Cetec-ERP/Cetec-Design-System/issues/4 . To-Dos:

- [x] Vertical Direction isn't rendering as CSS
- [x] Add "weight" prop for adjusting line thickness
- [x] ~~possibly refactor dive to take width x height and use a bg color. Right now weight of 1 is actually 2px, b/c it's the border of the div~~


atkinsdavid:
@Etwigg @Yash-CETEC @shaunrfox Hitting a bit of a wall here. 2 big issues that I'm running into currently:

1) passing direction as horizontal, or omitting it and using "default", is working as expected. But when you set the value to "vertical" none of the variant css properties get rendered, and the div doesn't get a class name for the direction variant
2) I need to figure out how to use NumericSizes tokens to pass into a prop called 'weight', and have that value apply to the 'borderWidth' in the recipe (or possibly later have that become part of the div WxH so that we aren't using border)

src/recipes/divider.ts
```diff
@@ -0,0 +1,43 @@
...
+  weight: {
+    thin: { '--divider-weight': 'sizes.1' },
+    medium: { '--divider-weight': 'sizes.2' },
+    bold: { '--divider-weight': 'sizes.4' },
+  },
+};
+
+export const dividerRecipe = defineRecipe({
+  className: 'divider',
+  jsx: ['Divider'],
```
shaunrfox, line 33: @atkinsdavid In this `jsx` property, 'divider' was not capitalized, which was somehow preventing all but the default classes from being rendered.

src/components/Divider/Divider.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { divider, type DividerVariantProps } from '@styled-system/recipes';
+
+import { cx } from '@styled-system/css';
+import { splitProps } from '~/utils/splitProps';
+
+export type DividerProps = Omit<BoxProps, keyof DividerVariantProps> &
+  DividerVariantProps & {
+    direction?: string;
+    weight?: string;
+  };
```
atkinsdavid, line 11: @shaunrfox I see that color isn't here. Are we still wanting to have that as a prop? I can take another crack at it with your changes if so, but didn't know if you'd rethought that while messing with this.

src/recipes/divider.ts
```diff
@@ -0,0 +1,43 @@
...
+  flexGrow: 0,
+  flexShrink: 0,
+};
+
+const dividerVariants = {
+  direction: {
+    horizontal: {
+      height: '1',
+      width: 'full',
+      borderTopWidth: 'var(--divider-weight)',
```
atkinsdavid, line 16: @shaunrfox Is going just one of the four borders the preferable option to you instead of trying to refactor this into a div with the given color? Definitely seems like a more straightforward way to solve the problem of the wight getting doubled up by the parallel borders anyway.

src/recipes/divider.ts
atkinsdavid: Interesting. Might have to skim through the Panda docs again, maybe there's something in there that would've called that out.

src/recipes/divider.ts
```diff
@@ -0,0 +1,43 @@
...
+    vertical: {
+      height: 'full',
+      width: '1',
+      borderLeftWidth: 'var(--divider-weight)',
+    },
+  },
+  weight: {
+    thin: { '--divider-weight': 'sizes.1' },
+    medium: { '--divider-weight': 'sizes.2' },
+    bold: { '--divider-weight': 'sizes.4' },
```
atkinsdavid: @shaunrfox I'd originally had these in there as placeholders as I was trying to get anything at all to work. Do we want to keep the pre-defined weight options, or allow weight to use any of the sizes tokens instead?

shaunrfox (approved):
Looks great!

Etwigg (approved):
noice
:shipit:

Issue #22, opened 2025-03-07
Radio


Etwigg:
[Radio Figma](https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=704-261&m=dev&t=zcVxkrV3CMOJ1qSw-1)

Issue #11, opened 2025-03-03
Component Refactors


Etwigg:
# Component refactors and changes go here.
[Components Figma Page](https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=0-1&m=dev&t=zcVxkrV3CMOJ1qSw-1)
Each component issue should have a link to it's figma page as well.

### Type Definitions
- Use BoxProps as a base if the component is based off of Box at all. BoxProps should include SystemStyleObject, so you don't need to extend that anywhere either. Omit the VariantProps from BoxProps and extend the VariantProps.
```typescript
export type CodeProps = Omit<BoxProps, keyof CodeVariantProps | keyof TextProps> & CodeVariantProps & TextProps & {
  children?: string | React.ReactNode;
  lang?: string;
};
```
In this example the Code componen will use Box and Text, so we need to extend both into the CodeProps definition and make sure CodeVariantProps is the source of truth.
- Make sure to export the type definition!
- Use any token types that are relevant such as in the Text component
```typescript
import type { FontToken, FontSizeToken, FontWeightToken } from '@styled-system/tokens';
export type TextProps = Omit<BoxProps, keyof TextVariantProps> & TextVariantProps & {
  italic?: boolean;
  family?: FontToken;
  bold?: boolean;
  underline?: boolean;
  size?: FontSizeToken;
  weight?: FontWeightToken;
  children?: string | React.ReactNode;
  as?: ElementType;
  className?: string;
};
```
### SplitCss
- Use the splitCss utility function to get className while passing all non-style props through to the component
```typescript
import { text, type TextVariantProps } from '@styled-system/recipes';
import { splitProps } from '~/utils/splitProps';

const [ className, otherProps ] = splitProps(props);
<Box
      as={as}
      className={cx(
        text({ family, bold, underline, italic, size, weight }),
        className as string,
      )}
      {...otherProps}
    >
      {children}
    </Box>
```
#### Import Order 
I would prefer when we import our own functions for them to be the last import.
Components and Type Declarations > Built In Functions > Utility Functions 

### Recipes
- Break recipes out into 'Base', 'Variant', then the recipe declaration.
```typescript
const textBase = {
};
const textVariants = {
};
export const textRecipe = defineRecipe({
  className: 'text',
  jsx: ['Text'],
  base: textBase,
  variants: textVariants,
  defaultVariants: {
  },
});
```
### Index File
[...]

PR #45, opened 2025-05-16
Radio Component

smitjp150622:
issue #22 

Create Radio Component

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 Cetec-ERP
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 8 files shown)


src/components/Code/Code.tsx
```diff
@@ -27,7 +27,7 @@ export const Code: React.FC<CodeProps> = (
       lang={lang}
       {...otherProps}
     >
-      <Text>
+      <Text color="slate.0">
```
Etwigg, line 30: This should be handled in the recipe?

src/components/Radio/index.tsx
```diff
@@ -0,0 +1 @@
+export {Radio} from "./Radio";
```
Etwigg: Export the type with the component!

Etwigg: Make sure you're following the conventions listed here: https://github.com/Cetec-ERP/Cetec-Design-System/issues/11

src/components/Code/Code.tsx
Etwigg: It's fine for now, the code component needs to be refactored later anyways..

Etwigg (changes requested):

src/components/Radio/Radio.tsx
Etwigg, line 1: Follow the guidelines here: https://github.com/Cetec-ERP/Cetec-Design-System/issues/11

smitjp150622: Yes @Etwigg I follow by the guidelines, but please let me know if I've missed anything.

Etwigg: Nevermind on the guidelines, I see you just used checkbox as a template, which is fine. 
However, the radio also doesn't un-select when testing. The selected/unselected behavior should function similarly to the checkbox.

smitjp150622: Since we're using radio buttons as the input type, they behave as a single-select option within a group.

For example, in a "Gender" group, the user should be able to select only one option — either "Male" or "Female", not both.

If we use "selected/unselected" terminology, it might confuse users and lead them to select multiple options, which could cause issues when storing the value.

I've also added a group for testing purposes — please check it on your end when you get a chance.

LICENSE.md
Etwigg, line 1: Why's this showing as a new file? There's already a license in main.

Etwigg (approved):

Etwigg (approved):
Ok, looks good.
I was thinking that we'd need to be able to deselect any given option, but I think that in that scenario (where we would want to allow the user to do that), we would just use a checkbox. I can't think of any examples of being able to deselect a radio.
:shipit:

PR #78, opened 2025-05-19
Checkbox component refactor

### LINKED ISSUE #19: CheckBox
[Checkbox Figma](https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=170-562&m=dev&t=zcVxkrV3CMOJ1qSw-1)

Double check CSS is correct w/ the figma ( should be good here, but just want to make sure we've got everything down.)

- [ ] Future: Build out Indeterminate and Error states (Don't start this yet. Look into it after the other components are done or we have a higher order component that needs to consume CheckBox).

Yash-CETEC:


src/components/CheckBox/CheckBox.tsx
Etwigg, line 1: I don't really understand these changes? I thought we were at a good spot already with this component?

src/recipes/checkbox.ts
```diff
@@ -1,88 +1,81 @@
...
+          opacity: '0.4',
+          pointerEvents: 'none',
+          cursor: 'none',
+        },
+        display: 'inline-grid',
+      },
+      _error: {
+        display: 'inline-grid',
+        '& ~ svg': {
+          fill: 'status.error.default',
```
Etwigg: This doesn't seem to be working.

src/styles/tokens.ts
```diff
@@ -447,6 +447,42 @@ export const breakpoints = {
   '2xl': '1536px',
 };
 
+export const opacity = {
```
Etwigg: why?

Etwigg (changes requested):
image

Error state doesn't seem to be working.

It looks like you're making a ton of extra changes that are not needed and are not part of the scope.

What's the reason for the extra stuff on this PR?

src/recipes/checkbox.ts
Yash-CETEC: The colour error in Figma is currently shown as `status.error.default`. Previously, this was consistent with our token naming; however, with the recent semantic token updates, it has now been revised to` error.default`. Updated according to that.

src/styles/tokens.ts
Yash-CETEC: Opacity is property for which values can be defined as tokens so added that for the same rather than adding value directly.

src/components/CheckBox/CheckBox.tsx
Yash-CETEC: The following new changes are implemented after the documentation is added to Figma.

=> Checkbox component changes:

1.  Addition of props for `error` and `disabled` for handling attribute `data-error` and `data-indeterminate`, and encapsulating both attributes in the checkbox component.
3. Removed the prop `children` and done the needful changes in component for the same.
4. Created `ExtraPropertyCheckBox` in `App.tsx` file so that the dynamic behaviour of `indeterminate` and `error` state can be tested on the page.
5. Previously, we were handling the `focus visible` event of the checkbox by modifying the outline when the component's focus visible state was active. But new Figma changes suggested `checkbox-focus` icon's use, so I replaced css approach with `checkbox-focus` icon.

=> Tokens Changes:
1. Panda config allows selected CSS properties to be defined as tokens. The `opacity` property was one of them, so I created `opacity` tokens and defined them in panda.config.js
2.  Replaced all the opacity values with tokens in the project.

P.S. Remaining changes are visible due to ESLint formatting. Please let me know if any of the above change you would like me to remove.

src/App.tsx
```diff
@@ -134,6 +134,36 @@ const Header: React.FC = () => {
...
+  return (
+    <CheckBox
+      checked={isChecked}
+      // there will be some conditions based on states and/or props which will dynamically add or remove the prop
+      // and which in turn will make error or indeterminate state active or deactive
+      {...propertyProps}
+      onClick={handleChange}
+    />
+  );
+};
```
Etwigg: Why are we doing all of this? This is just an example of what you'd want to do in the checkbox component file, right?

Etwigg: data-error doesn't work?

Etwigg (changes requested):
I want to revisit the opacity tokens... I'm not sure we need these, and I know that having the token be the same as the actual value doesn't make sense.
If there are a couple opacity values we use everywhere, then I can see making some tokens for them, but they'll need semantic names.
@shaunrfox thoughts?

src/App.tsx
Yash-CETEC: I created a working version for indefinite and error section but which will show error and indefinite state at first but on click can be changed. This was done in the last pr and was already approved this time only eslint changes and some other minor changes were done. But only static version are needed to show, I can remove this. Please let me know your preference for the same.

Etwigg:
@shaunrfox 
Bump

Etwigg:
@Yash-CETEC 
Ditch the opacity tokens for now, please!
Thank you

Yash-CETEC:
> @Yash-CETEC Ditch the opacity tokens for now, please! Thank you

Done! @Etwigg

src/App.tsx
```diff
@@ -134,6 +134,36 @@ const Header: React.FC = () => {
...
+
+  return (
+    <CheckBox
+      checked={isChecked}
+      // there will be some conditions based on states and/or props which will dynamically add or remove the prop
+      // and which in turn will make error or indeterminate state active or deactive
+      {...propertyProps}
+      onClick={handleChange}
+    />
+  );
```
Etwigg: We're not going to do it this way. See shaun's comments in the chat and on Smit's PR for how we want to handle error state now.

Yash-CETEC: This is a wrapper component that demonstrates how the Checkbox will work with error and indeterminate props. The actual implementation is inside the checkbox component, where you can see it is working as per mentioned by Shaun.

Screenshot 2025-06-04 at 2 23 04 PM

Etwigg: I understand that. You wrote this as a suggestion, so I was letting you know that's not the approach we'll be taking and you can remove it.

src/App.tsx
Etwigg:
```suggestion
```

src/App.tsx
```diff
@@ -134,6 +134,36 @@ const Header: React.FC = () => {
   );
 };
 
+// This approach is created to show actuall working of checkbox in react page.
+// This is a very basic example created to show how indeterminate or error state
+// can be removed once checkbox is clicked.
+// but in real life scenario the isPropertyRequired state will be replaced with
+// some other states or props or combination of both which will allow us
+// dynamic handling of states.
```
Etwigg:
```suggestion
```

src/App.tsx
```diff
@@ -369,28 +399,27 @@ const AppContent: React.FC = () => {
           <Section>
             <Heading level="h2">Checkboxes</Heading>
             <HStack gap={'40'} alignItems={'flex-end'}>
-              <CheckBox>asdf</CheckBox>
+              <CheckBox />
               <CheckBox defaultChecked={true} />
-              <CheckBox data-indeterminate={true} />
+              <ExtraPropertyCheckBox property="indeterminate" />
```
Etwigg:
```suggestion
```

src/App.tsx
```diff
@@ -369,28 +399,27 @@ const AppContent: React.FC = () => {
...
               <CheckBox disabled />
-              <CheckBox data-error={true} />
+              <ExtraPropertyCheckBox property="error" />
```
Etwigg:
```suggestion
```

src/App.tsx
```diff
@@ -369,28 +399,27 @@ const AppContent: React.FC = () => {
           <Section>
             <Heading level="h2">Checkboxes</Heading>
             <HStack gap={'40'} alignItems={'flex-end'}>
-              <CheckBox>asdf</CheckBox>
+              <CheckBox />
               <CheckBox defaultChecked={true} />
```
Etwigg, line 374:
```suggestion
              <CheckBox defaultChecked={true} />
              <CheckBox indeterminate={true} />
              <CheckBox error={true} />
              <CheckBox disabled />
```

Etwigg (dismissed):

Issue #26, opened 2025-03-31
TextInput


Etwigg:
Figma: https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=111-255&t=KFmha8hViGRbWGH4-1

Ignore the 'Form Field' hanging out in that Figma link. That's for later. This is just the TextInput component and it's styles.

PR #79, opened 2025-05-19
Create text input component

smitjp150622:
Issue #26

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -22,7 +22,7 @@ export const TextInput: React.FC<TextInputProps> = (
     return(
         <Box
         as="input"
-        data-error={error ? '' : undefined} 
+        {...error && { 'data-error': true }}
         className={cx(
             textinput({size, autoSize}),
             className
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -11,6 +11,7 @@ export type TextInputProps = Omit<BoxProps, keyof TextinputVariantProps> & Texti
 
 export const TextInput: React.FC<TextInputProps> = (
     {
+        size,
         error,
         ...props
     }: TextInputProps
@@ -21,7 +22,7 @@ export const TextInput: React.FC<TextInputProps> = (
         <Box
         as="input"
         className={cx(
-            textinput({}),
+            textinput({size}),
             className
         )}
         {...otherProps}
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -1,0 +1,31 @@
+import { splitProps } from '~/utils/splitProps';
+import {Box, type BoxProps} from '../Box/Box';
+import {textinput, type TextinputVariantProps } from '@styled-system/recipes';
+import { cx } from '@styled-system/css';
+
+
+export type TextInputProps = Omit<BoxProps, keyof TextinputVariantProps> & TextinputVariantProps &{
+    size?: 'small' | 'medium' | 'large';
+    className?: string;
+}
+
+export const TextInput: React.FC<TextInputProps> = (
+    {
+        size,
+        ...props
+    }: TextInputProps
+) => {
+
+    const [ className, otherProps] = splitProps(props);
+    return(
+        <Box
+        as="input"
+        className={cx(
+            textinput({size}),
+            className
+        )}
+        {...otherProps}
+        />
+    )
+
+}
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -7,12 +7,11 @@ import { cx } from '@styled-system/css';
 export type TextInputProps = Omit<BoxProps, keyof TextinputVariantProps> & TextinputVariantProps &{
     disabled?: boolean,
     error?: boolean
-
 }
 
 export const TextInput: React.FC<TextInputProps> = (
     {
-        size,
+        error,
         ...props
     }: TextInputProps
 ) => {
@@ -22,7 +21,7 @@ export const TextInput: React.FC<TextInputProps> = (
         <Box
         as="input"
         className={cx(
-            textinput({size}),
+            textinput({}),
             className
         )}
         {...otherProps}
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -5,7 +5,6 @@ import { cx } from '@styled-system/css';
 
 
 export type TextInputProps = Omit<BoxProps, keyof TextinputVariantProps> & TextinputVariantProps &{
-    disabled?: boolean,
     error?: boolean,
 }
 
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -5,8 +5,9 @@ import { cx } from '@styled-system/css';
 
 
 export type TextInputProps = Omit<BoxProps, keyof TextinputVariantProps> & TextinputVariantProps &{
-    size?: 'small' | 'medium' | 'large';
-    className?: string;
+    disabled?: boolean,
+    error?: boolean
+
 }
 
 export const TextInput: React.FC<TextInputProps> = (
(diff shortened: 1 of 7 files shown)


src/components/Textarea/Textarea.tsx
```diff
@@ -24,7 +24,7 @@ export const Textarea: React.FC<TextareaProps> = (
       <Text as='span'>{label}</Text>
       <Box
         as="textarea"
-        id={props.id}
+        id={props.id  }
```
Etwigg:
```suggestion
        id={props.id}
```

Etwigg (changes requested):
It looks like you're using the 'focused-visible' variant on all of the focus states?
Shaun replied to your question in the chat, maybe that's just what was missing.

Looks pretty good otherwise!

src/components/Textarea/Textarea.tsx
smitjp150622: Hello, @Etwigg I changed the focus according to Shaun's advice. and remove the blank space in the Textarea.tsx file.

Etwigg: looks like it's still happening?
There should be two focus states. The first is when you click in (normal focus). The second is "tab" focus.
image

Etwigg: Unless I'm mistaken @shaunrfox

smitjp150622: I'm using both `:focus` and `:focus-visible states`, but when I click on the input or navigate using the keyboard, only `:focus-visible` seems to be applied.

Etwigg:
@smitjp150622 
Let me know when this is ready to review!
Thanks

smitjp150622:
@Etwigg Please review this PR, and shaun already pinged me we only setting up `:focus-visible`.

shaunrfox (changes requested):
This should be treated very similarly to the TextArea component. Please apply all the changes I mentioned in https://github.com/Cetec-ERP/Cetec-Design-System/pull/81

src/recipes/textinput.ts
```diff
@@ -0,0 +1,69 @@
...
+  },
+};
+
+const textInputBase = {
+  position: 'relative',
+  borderWidth: '1',
+  borderColor: 'slate.30',
+  borderStyle: 'solid',
+  borderRadius: '4',
+  fontFamily: 'mono',
```
shaunrfox: This should be `fontFamily: 'body'`

shaunrfox (changes requested):
Also apply the feedback I added in the textArea PR

src/recipes/textinput.ts
shaunrfox: Also, the heights of the inputs are incorrect. Need to add `lineHeight: 'default'`

Etwigg:
Error state focus issues that Shaun already pointed out:
image

Is default filled supposed to be uneditable?

No, it's not actually uneditable. I used the content prop to set the value, so while it is displayed, it's not editable. Also, since we haven’t added any onChange functionality, editing isn't working.

src/recipes/textinput.ts
```diff
@@ -47,12 +52,14 @@ const textInputBase = {
   },
 
   _focus: {
+    outlineColor: { base: 'slate.90', _dark: 'slate.0' },
     borderColor: { base: 'slate.90', _dark: 'slate.0' },
   },
-  _error: {
+  '&[data-error="true"]': {
```
Etwigg: We want to use/implement _error here. We don't need to assign data-error a value, just check for the existence of data-error.

Etwigg (changes requested):

src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,34 @@
...
+        autoSize = false,
+        ...props
+    }: TextInputProps
+) => {
+
+    const [ className, otherProps] = splitProps(props);
+    return(
+        <Box
+        as="input"
+        data-error={error ? '' : undefined}
```
Etwigg: Change this to check for error and if it exists, add data-error attribute
```suggestion
        error ? data-error : undefined
```

Etwigg (changes requested):

smitjp150622:
@Etwigg and @shaunrfox, 
I investigated the data-error issue and found that even when `data-error="false"`, the error recipe styles are still being applied. This is because in our conditions.ts file, we're currently using the selector
**error: '&:is([data-error], [error])'**

This matches both `data-error="true"` and` data-error="false"` (or even just data-error without a value), which causes incorrect styling.

To fix this, we need to update the condition to target only `data-error="true"`, like this:
**error: '&:is([data-error=true], [error])'**

This change would make the error styles apply only when the error state is explicitly true, and prevent unintended styling. I verified that even our older components are affected by the current behavior.

Alternatively, another way to handle this is to directly update the recipe files and use:
**_error: to '&[data-error="true"]'**

Please suggest which approach we should move forward with:
1. Update the global condition in conditions.ts
2.Update the individual recipe files with the stricter selector

Screenshot 2025-06-03 at 11 28 47 AM


I recommend option 1 for consistency and ease of maintenance across all components.

Etwigg:
Putting Shaun's reply here for posterity:

Hi @Smit Patel . Yes, this is what I was [referring to in my comments](https://github.com/Cetec-ERP/Cetec-Design-System/pull/81#issuecomment-2913208432). 

I want to ensure two things as we go forward:
We don't alter the conditions that came with Panda
We use the conditions in our recipes wherever possible

This means that our recipes should leverage the existing conditions as documented, and that we can make changes to our components to better align with these conditions.

The _error shorthand condition references these selectors: error: '&:is([data-error], [error])'.

So far, in our components we've been setting the error like this:
❌ data-error={error ? '' : undefined}

This can lead to both 'true' and 'false' showing the error state styles.

I think a safer, more reliable approach would be to forget setting true/false/undefined, and instead don't even render data-error unless the error state exists:
✅ {...error && { 'data-error': true }}

This should actually be our convention for most (or all) of the boolean conditions that we're using in our components.

Examples:
// disabled
❌ disabled={disabled}
✅ {...disabled && { 'data-disabled': true }}

// checkbox indeterminate state
✅ {...indeterminate && { 'data-indeterminate': true }}

// required
✅ {...required & { 'data-required': true }}

Yash-CETEC:
> Putting Shaun's reply here for posterity:
> 
> Hi @Smit Patel . Yes, this is what I was [referring to in my comments](https://github.com/Cetec-ERP/Cetec-Design-System/pull/81#issuecomment-2913208432).
> 
> I want to ensure two things as we go forward: We don't alter the conditions that came with Panda We use the conditions in our recipes wherever possible
> 
> This means that our recipes should leverage the existing conditions as documented, and that we can make changes to our components to better align with these conditions.
> 
> The _error shorthand condition references these selectors: error: '&:is([data-error], [error])'.
> 
> So far, in our components we've been setting the error like this: ❌ data-error={error ? '' : undefined}
> 
> This can lead to both 'true' and 'false' showing the error state styles.
> 
> I think a safer, more reliable approach would be to forget setting true/false/undefined, and instead don't even render data-error unless the error state exists: ✅ {...error && { 'data-error': true }}
> 
> This should actually be our convention for most (or all) of the boolean conditions that we're using in our components.
> 
> Examples: // disabled ❌ disabled={disabled} ✅ {...disabled && { 'data-disabled': true }}
> 
> // checkbox indeterminate state ✅ {...indeterminate && { 'data-indeterminate': true }}
> 
> // required ✅ {...required & { 'data-required': true }}

Required and disabled are the attributes which are already provided by the input tag itself `data-` format is generally used to add custom data attributes so for cases like indeterminate & error the suggestion works but for properties like disabled and required disabled={disabled} and required={required} can be used

Even this can be avoided page where component is used can directly add `disabled` or `required` and spreading of prop will automatically handle this.

src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,34 @@
+import { splitProps } from '~/utils/splitProps';
+import {Box, type BoxProps} from '../Box/Box';
+import {textinput, type TextinputVariantProps } from '@styled-system/recipes';
+import { cx } from '@styled-system/css';
+
+
+export type TextInputProps = Omit<BoxProps, keyof TextinputVariantProps> & TextinputVariantProps &{
+    disabled?: boolean,
```
Etwigg:
```suggestion
```

Etwigg: We don't need disabled here, right? It's automatically handled.

Etwigg (dismissed):

Issue #48, opened 2025-05-16
TextArea


Etwigg:
https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1395-6827&m=dev&t=zorehhi4oWHDjOBw-1

smitjp150622:
@Etwigg Please check PR [TextArea Component](https://github.com/Cetec-ERP/Cetec-Design-System/pull/81)

PR #81, opened 2025-05-20
Refactor Textarea Compnent

smitjp150622:
TextArea #48

### DIFF
--- a/src/recipes/textarea.ts
+++ b/src/recipes/textarea.ts
@@ -12,9 +12,12 @@ const textareaBase = {
   outlineColor: 'transparent',
   outlineOffset: '1',
   _focusVisible: {
-    borderColor: 'slate.90',
+    outlineWidth: '2',
     outlineColor: { base: 'slate.90', _dark: 'slate.0' },
-  },  
+  },
+  _focus: {
+    borderColor: 'slate.90',
+  },
   _error: {
     display: 'inline-grid',
     borderColor: 'red.50',
@@ -31,24 +34,24 @@ const textareaBase = {
 };
 
 const textareaVariants = {
-  size:{
-    small:{
-      px:'8',
-      minHeight:'48',
-      fontSize:'14',
+  size: {
+    small: {
+      px: '8',
+      minHeight: '48',
+      fontSize: '14',
     },
-    medium:{
-      py:'3',
-      px:'10',
-      fontSize:'16',
-      minHeight:'64',
+    medium: {
+      py: '3',
+      px: '10',
+      fontSize: '16',
+      minHeight: '64',
     },
-    large:{
+    large: {
       py: '7',
       px: '12',
       fontSize: '16',
-      minHeight:'80',
-    }
+      minHeight: '80',
+    },
   },
 };
 
--- a/src/recipes/textarea.ts
+++ b/src/recipes/textarea.ts
@@ -16,7 +16,7 @@ const textareaBase = {
     outlineColor: { base: 'slate.90', _dark: 'slate.0' },
     borderColor: { base: 'slate.90', _dark: 'slate.0' },
   },
-  '&[data-error="true"]': {
+  _error: {
     display: 'inline-grid',
     borderColor: 'error.default',
     _focus: {
--- a/src/recipes/textarea.ts
+++ b/src/recipes/textarea.ts
@@ -7,17 +7,15 @@ const textareaBase = {
   borderStyle: 'solid',
   borderColor: 'slate.30',
   borderRadius: '4',
-  outlineWidth: '2',
-  outlineStyle: 'solid',
-  outlineColor: 'transparent',
-  outlineOffset: '1',
-  _focusVisible: {
-    outlineWidth: '2',
-    outlineColor: { base: 'slate.90', _dark: 'slate.0' },
+  _focus:{
+    borderColor: {base: 'slate.90', _dark: 'slate.0'}
   },
   _error: {
     display: 'inline-grid',
-    borderColor: 'red.50',
+    borderColor: 'error.default',
+    _focus:{
+      borderColor: 'error.default',
+    }
   },
   _placeholder: {
     color: {
@@ -33,27 +31,32 @@ const textareaBase = {
 const textareaVariants = {
   size: {
     small: {
-      py: '3',
+      py: '0',
       px: '8',
       minHeight: '48',
       fontSize: '14',
-      lineHeights: '14',
+      lineHeight: 'tight',
     },
     medium: {
       py: '3',
       px: '10',
       fontSize: '16',
       minHeight: '64',
-      lineHeights: '16',
+      lineHeight: 'numericSizes.16',
     },
     large: {
       py: '7',
       px: '12',
       fontSize: '16',
       minHeight: '80',
-      lineHeights: '16',
+      lineHeight: 'numericSizes.16',
     },
   },
+  autoSize: {
+    true: {
+      fieldSizing: 'content'
+    }
+  }
 };
 
 export const textareaRecipe = defineRecipe({
--- a/src/recipes/textarea.ts
+++ b/src/recipes/textarea.ts
@@ -15,9 +15,6 @@ const textareaBase = {
     outlineWidth: '2',
     outlineColor: { base: 'slate.90', _dark: 'slate.0' },
   },
-  _focus: {
-    borderColor: 'slate.90',
-  },
   _error: {
     display: 'inline-grid',
     borderColor: 'red.50',
@@ -36,21 +33,25 @@ const textareaBase = {
 const textareaVariants = {
   size: {
     small: {
+      py: '3',
       px: '8',
       minHeight: '48',
       fontSize: '14',
+      lineHeights: '14',
     },
     medium: {
       py: '3',
       px: '10',
       fontSize: '16',
       minHeight: '64',
+      lineHeights: '16',
     },
     large: {
       py: '7',
       px: '12',
       fontSize: '16',
       minHeight: '80',
+      lineHeights: '16',
     },
   },
 };
--- a/src/recipes/textarea.ts
+++ b/src/recipes/textarea.ts
@@ -26,6 +26,9 @@ const textareaBase = {
   _disabled: {
     opacity: 0.4,
   },
+  '&[data-error]:focus': {
+    borderColor: 'error.default',
+  },
 };
 
 const textareaVariants = {
--- a/src/recipes/textarea.ts
+++ b/src/recipes/textarea.ts
@@ -2,67 +2,53 @@ import { defineRecipe } from '@pandacss/dev';
 
 const textareaBase = {
   position: 'relative',
-  display: 'inline-grid',
-  verticalAlign: 'top',
-  alignItems: 'center',
-  padding: '0',
-  fontFamily: 'body',
-  fontSize: '16',
-  fontWeight: 'normal',
-  lineHeight: 'normal',
-  color: { base: 'slate.90', _osDark: 'slate.5' },
-  borderRadius: '8',
-  width: 'full',
-  '&::after, & textarea': {
-    width: 'auto',
-    minWidth: '16',
-    maxWidth: 'full',
-    font: 'inherit',
-    py: '4',
-    px: '8',
-    m: '0',
-    resize: 'none',
-    appearance: 'none',
-    borderWidth: '1',
-    borderStyle: 'solid',
-    borderRadius: '4',
-    borderColor: 'transparent',
-    background: { base: 'slate.0', _osDark: 'slate.90' },
-    _hover: {
-      borderColor: { base: 'slate.20', _osDark: 'slate.30' },
-    },
-    _focus: {
-      borderColor: { base: 'slate.90', _osDark: 'slate.5' },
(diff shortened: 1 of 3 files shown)


Etwigg (changes requested):
Your focus and focus-visible states are the same. 
And the focus-visible is supposed to be a border around the normal element:
image
vs.
image

This is especially apparent for the error state:
image

Placeholder text is incorrectly positioned:
image

Vs.

image

You need a bit more space between the cursor and border in the above screenshots.

smitjp150622:
Screenshot 2025-05-22 at 4 54 40 PM


@Etwigg I've already applied the styling based on the Figma design. According to the design, the small-sized text area component should not have any vertical padding.

If padding is required, please let me know the exact pixel values so I can update it accordingly.
I'll review the remaining points as well.

Figma link:- [Text Area](https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1396-7347&m=dev)

Etwigg:
> Screenshot 2025-05-22 at 4 54 40 PM
> @Etwigg I've already applied the styling based on the Figma design. According to the design, the small-sized text area component should not have any vertical padding.
> 
> If padding is required, please let me know the exact pixel values so I can update it accordingly. I'll review the remaining points as well.
> 
> Figma link:- [Text Area](https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1396-7347&m=dev)

Maybe it's a difference in line height of the placeholder?
When you look at the variants, the cursor has a 3pd vertical padding:
image

src/recipes/textarea.ts
```diff
@@ -2,67 +2,57 @@ import { defineRecipe } from '@pandacss/dev';
...
+  outlineStyle: 'solid',
+  outlineColor: 'transparent',
+  outlineOffset: '1',
+  _focusVisible: {
+    outlineWidth: '2',
+    outlineColor: { base: 'slate.90', _dark: 'slate.0' },
+  },
+  _error: {
+    display: 'inline-grid',
+    borderColor: 'red.50',
```
shaunrfox: This should use our `error.default` token

src/recipes/textarea.ts
```diff
@@ -2,67 +2,57 @@ import { defineRecipe } from '@pandacss/dev';
...
+  color: { base: 'slate.90', _dark: 'slate.0' },
+  borderWidth: '1',
+  borderStyle: 'solid',
+  borderColor: 'slate.30',
+  borderRadius: '4',
+  outlineWidth: '2',
+  outlineStyle: 'solid',
+  outlineColor: 'transparent',
+  outlineOffset: '1',
+  _focusVisible: {
```
shaunrfox: I've removed the `:focus-visible` state in Figma. Let's just change this to `_focus` with this:
```
_focus: {
  borderColor: { base: 'slate.90', _dark: 'slate.0' },
},
```

src/recipes/textarea.ts
```diff
@@ -2,67 +2,57 @@ import { defineRecipe } from '@pandacss/dev';
...
-      borderColor: { base: 'slate.20', _osDark: 'slate.30' },
-    },
-    _focus: {
-      borderColor: { base: 'slate.90', _osDark: 'slate.5' },
+  color: { base: 'slate.90', _dark: 'slate.0' },
+  borderWidth: '1',
+  borderStyle: 'solid',
+  borderColor: 'slate.30',
+  borderRadius: '4',
+  outlineWidth: '2',
```
shaunrfox: These outline styles can now be removed

src/recipes/textarea.ts
```diff
@@ -2,67 +2,57 @@ import { defineRecipe } from '@pandacss/dev';
...
+  outlineStyle: 'solid',
+  outlineColor: 'transparent',
+  outlineOffset: '1',
+  _focusVisible: {
+    outlineWidth: '2',
+    outlineColor: { base: 'slate.90', _dark: 'slate.0' },
+  },
+  _error: {
```
shaunrfox, line 20: When the textarea is both `_error` and `_focus` the error state should remain visible until the prop is removed. This probably just needs another declaration for this combined state.

src/recipes/textarea.ts
```diff
@@ -2,67 +2,57 @@ import { defineRecipe } from '@pandacss/dev';
...
-  gridTemplateColumns: 'auto 1fr',
 };
 
 const textareaVariants = {
-  variant: {
-    autoGrow: {
-      width: 'fit-content',
+  size: {
+    small: {
+      py: '3',
```
shaunrfox: The padding values in these size variants should match what we're doing for Button (not IconButton). So this should be `py: '0'`.

src/recipes/textarea.ts
```diff
@@ -2,67 +2,57 @@ import { defineRecipe } from '@pandacss/dev';
...
+      px: '8',
+      minHeight: '48',
+      fontSize: '14',
+      lineHeights: '14',
```
shaunrfox: These line-height declarations aren't currently being applied because the property should not have an 's' on the end. They also need to be changed to `lineHeight: 'tight',` for all size variants.

shaunrfox (changes requested):
In addition to my other comments, I'd like us to consider adding another prop. @Etwigg I'd like to get your take on this.

There's a partially supported CSS property for text inputs and textareas that allows the input to grow based on the content. We wouldn't want this in all situations, but it would be nice to have it available in certain cases. 

Since it's not fully supported, it can be considered a 'progressive enhancement' and should have no detrimental effects in unsupported browsers.

MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing

Here's what I propose:

- A new boolean prop called `autoSize` (defaulted to 'false')
- The CSS for this would be `fieldSizing: 'content'`

What do y'all think?

shaunrfox:
> > Screenshot 2025-05-22 at 4 54 40 PM
> > @Etwigg I've already applied the styling based on the Figma design. According to the design, the small-sized text area component should not have any vertical padding.
> > If padding is required, please let me know the exact pixel values so I can update it accordingly. I'll review the remaining points as well.
> > Figma link:- [Text Area](https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1396-7347&m=dev)
> 
> Maybe it's a difference in line height of the placeholder? When you look at the variants, the cursor has a 3pd vertical padding: image

Sorry, I just saw this. This is not the correct way to measure the padding in Figma for this component. That cursor is just a random size. The comments I just added should be the right way to go.

Etwigg:
> In addition to my other comments, I'd like us to consider adding another prop. @Etwigg I'd like to get your take on this.
> 
> There's a partially supported CSS property for text inputs and textareas that allows the input to grow based on the content. We wouldn't want this in all situations, but it would be nice to have it available in certain cases.
> 
> Since it's not fully supported, it can be considered a 'progressive enhancement' and should have no detrimental effects in unsupported browsers.
> 
> MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing
> 
> Here's what I propose:
> 
> * A new boolean prop called `autoSize` (defaulted to 'false')
> * The CSS for this would be `fieldSizing: 'content'`
> 
> What do y'all think?

Sure!

atkinsdavid:
> > What do y'all think?
> 
> Sure!

Agreed, I think there are plenty of valid situations where this would be a nice to have from the user perspective.

src/recipes/textarea.ts
smitjp150622: Hi @shaunrfox, I will change it, please update on figma also both the component text area and text input, and also additional prop also adding in figma.

shaunrfox:
Alright, this is looking pretty good. I just have a few comments:

## Focus state

In the current state of the code, the focus state is visually changing the border from 1px to 2px. This was supposed to just remain 1px, but it's inheriting styles from our globalStyles.

BUT I actually think this change is a better interaction. So let's keep it, but make a few changes so that it's explicit.

Add these to the recipe:

```JSX
const textareaBase = {
  // ...
  outlineWidth: '1',
  outlineStyle: 'solid',
  outlineColor: 'transparent',
  // ...
  _focus: {
    // ...
    outlineColor: { base: 'slate.90', _dark: 'slate.0' },
  }
}
```

## Error state

You added the CSS selector `'&[data-error]:focus'` for the combined error + focus states. This is not quite precise enough, and could cause the error state to appear by accident if we ever set `data-error="false"`:

Screenshot 2025-05-27 at 09 42 02 AM

Adding the following should be sufficient:

```JSX
_error: {
  // ...
  _focus: {
  // ...
    borderColor: 'error.default',
  },
},
```

## Field width

Lastly, this isn't necessarily apparent in the mockups, but all inputs (and most components) should grow to the width of their parent container. 

So we just need to add this to the base recipe: `width: 'full'`

Etwigg:
Error state doesn't look right at all:
image

The indicator here is still weird as well: 
image

smitjp150622:
> Error state doesn't look right at all: image
> 
> The indicator here is still weird as well: image


@Etwigg, I forgot to check the error focus in dark mode, but that issue is now resolved. However, regarding the indicator issue, could you please elaborate on the specific problems? I'm a bit confused and would appreciate more details to better understand and address them.

Etwigg:
> > Error state doesn't look right at all: image
> > The indicator here is still weird as well: image
> 
> @Etwigg, I forgot to check the error focus in dark mode, but that issue is now resolved. However, regarding the indicator issue, could you please elaborate on the specific problems? I'm a bit confused and would appreciate more details to better understand and address them.

I talked with Shaun, I'm wrong about the indicator! Disregard.

src/recipes/textarea.ts
```diff
@@ -2,21 +2,27 @@ import { defineRecipe } from '@pandacss/dev';
...
-    borderColor: {base: 'slate.90', _dark: 'slate.0'},
+  outlineWidth: '1',
+  outlineStyle: 'solid',
+  outlineColor: 'transparent',
+  _focus: {
+    outlineColor: { base: 'slate.90', _dark: 'slate.0' },
+    borderColor: { base: 'slate.90', _dark: 'slate.0' },
   },
-  _error: {
+  '&[data-error="true"]': {
```
Etwigg: Same thing here as mentioned in the other PRs

Etwigg (changes requested):

src/recipes/textarea.ts
```diff
@@ -2,21 +2,27 @@ import { defineRecipe } from '@pandacss/dev';
...
   borderRadius: '4',
   lineHeight: 'tight',
-  _focus:{
-    borderColor: {base: 'slate.90', _dark: 'slate.0'},
+  outlineWidth: '1',
+  outlineStyle: 'solid',
+  outlineColor: 'transparent',
+  _focus: {
+    outlineColor: { base: 'slate.90', _dark: 'slate.0' },
+    borderColor: { base: 'slate.90', _dark: 'slate.0' },
```
Etwigg, line 18: I don't think you're supposed to have both of these. This isn't what @shaunrfox was describing.
@shaunrfox 
Can you take a look at this?
Thanks

Etwigg (changes requested):

src/recipes/textarea.ts
shaunrfox: No, @Etwigg , that's correct. Here's what the final `textareaBase` styling should be:

```JSX
const textareaBase = {
  position: 'relative',
  width: 'full',
  color: { base: 'slate.90', _dark: 'slate.0' },
  borderWidth: '1',
  borderStyle: 'solid',
  borderColor: 'slate.30',
  borderRadius: '4',
  lineHeight: 'tight',
  outlineWidth: '1',
  outlineStyle: 'solid',
  outlineColor: 'transparent',
  _focus: {
    borderColor: { base: 'slate.90', _dark: 'slate.0' },
    outlineColor: { base: 'slate.90', _dark: 'slate.0' },
  },
  _error: {
    display: 'inline-grid',
    borderColor: 'error.default',
    _focus: {
      borderColor: { base: 'error.default', _dark: 'error.default' },
      outlineColor: { base: 'error.default', _dark: 'error.default' },
    },
  },
  _placeholder: {
    color: {
      base: 'slate.50',
      _dark: 'slate.40',
    },
  },
  _disabled: {
    opacity: 0.4,
  },
};
```

Etwigg (approved):

Etwigg (dismissed):
gucci

src/recipes/textarea.ts
```diff
@@ -2,67 +2,67 @@ import { defineRecipe } from '@pandacss/dev';
...
-    },
+  color: { base: 'slate.90', _dark: 'slate.0' },
+  borderWidth: '1',
+  borderStyle: 'solid',
+  borderColor: 'slate.30',
+  borderRadius: '4',
+  lineHeight: 'tight',
+  outlineWidth: '1',
+  outlineStyle: 'solid',
+  outlineColor: 'transparent',
```
Etwigg, line 14:
```suggestion
  outlineColor: 'transparent',
  resize:'both';
```

Etwigg: Resize was getting set to vertical somewhere? Odd.

Etwigg (approved):

Issue #46, opened 2025-05-16
Toggle


Etwigg:
Just the toggle component, not the full ToggleInput

https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1390-4813&m=dev&t=zorehhi4oWHDjOBw-1

Etwigg:
Previously 'Switch'!!

PR #82, opened 2025-05-26
Create switch component.

smitjp150622:
issue #46 

@Etwigg and @shaunrfox, I added the <label> element because I'm creating a switch component using an input of type checkbox. To customise the checkbox and style it like a switch, I needed a wrapper element—so I used a <label> as the container. This allows me to apply custom styles around the input. Additionally, I use an onChange function to handle switch toggling (select/unselect) and apply the corresponding styling dynamically.

### DIFF
--- a/panda.config.ts
+++ b/panda.config.ts
@@ -25,7 +25,7 @@ import {
   codeRecipe,
   boxRecipe,
   radioRecipe,
-  switchRecipe,
+  toggleRecipe,
 } from './src/recipes/index';
 
 // https://panda-css.com/docs/concepts/extend#removing-something-from-the-base-presets
@@ -120,7 +120,7 @@ export default defineConfig({
         pre: preRecipe,
         spinner: spinnerRecipe,
         divider: dividerRecipe,
-        switchbox: switchRecipe,
+        toggle: toggleRecipe,
         box: boxRecipe,
       },
       slotRecipes: {
--- a/panda.config.ts
+++ b/panda.config.ts
@@ -24,7 +24,8 @@ import {
   preRecipe,
   codeRecipe,
   boxRecipe,
-  radioRecipe
+  radioRecipe,
+  switchRecipe,
 } from './src/recipes/index';
 
 // https://panda-css.com/docs/concepts/extend#removing-something-from-the-base-presets
@@ -119,6 +120,8 @@ export default defineConfig({
         pre: preRecipe,
         spinner: spinnerRecipe,
         divider: dividerRecipe,
+        switchbox: switchRecipe,
+        //Panda might not like this since they already have a box
         box: boxRecipe,
       },
       slotRecipes: {
--- a/panda.config.ts
+++ b/panda.config.ts
@@ -121,7 +121,6 @@ export default defineConfig({
         spinner: spinnerRecipe,
         divider: dividerRecipe,
         switchbox: switchRecipe,
-        //Panda might not like this since they already have a box
         box: boxRecipe,
       },
       slotRecipes: {
--- a/panda.config.ts
+++ b/panda.config.ts
@@ -6,7 +6,7 @@ import {
 import pandaBasePreset from '@pandacss/preset-base';
 
 import * as tokens from './src/styles/tokens';
-import * as semanticTokens from "./src/styles/semanticTokens";
+import * as semanticTokens from './src/styles/semanticTokens';
 import { globalCss } from './src/styles/globalStyle';
 import { conditions } from './src/styles/conditions';
 import {
@@ -26,6 +26,7 @@ import {
   boxRecipe,
   radioRecipe,
   toggleRecipe,
+  textinputRecipe,
 } from './src/recipes/index';
 
 // https://panda-css.com/docs/concepts/extend#removing-something-from-the-base-presets
@@ -115,6 +116,7 @@ export default defineConfig({
         button: buttonRecipe,
         iconButton: iconButtonRecipe,
         input: inputRecipe,
+        textinput: textinputRecipe,
         textarea: textareaRecipe,
         code: codeRecipe,
         pre: preRecipe,
--- a/src/components/Switch/Switch.tsx
+++ b/src/components/Switch/Switch.tsx
@@ -20,14 +20,14 @@ export const Switch: React.FC<SwitchBoxProps> = ({
   return (
     <Label
       className={container}
-      data-disabled={disabled}
+      {...disabled && { 'data-disabled': true }}
     >
       <Box
         as="input"
         type="checkbox"
         className={input}
         data-checked={checked}
-        data-error={error}
+        {...error && { 'data-error': true }}
         {...props}
       />
       <Box
--- a/src/components/Switch/Switch.tsx
+++ b/src/components/Switch/Switch.tsx
@@ -28,13 +28,12 @@ export const Switch: React.FC<SwitchBoxProps> = ({
         className={input}
         {...props}
         data-checked={checked}
+        data-error={error}
       />
       <Box
         as="span"
         class={background}
-        data-checked={checked}
         name={'switch-bg'}
-        data-error={error}
       />
       <Icon name={'circle'} className={indicator} />
       <Icon name={'circle-check'} className={indicator} />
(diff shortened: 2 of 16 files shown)


Etwigg (changes requested):
The animation of the checkbox being checked seems faster than the codepen provided on the figma.

Etwigg:
@smitjp150622 
It's still not like the codepen: https://codepen.io/shaunrfox/pen/azzLVgp?editors=1100
It's not smooth at all.

https://github.com/user-attachments/assets/7346610b-e81d-44ed-b19f-0cf0332780fe

https://github.com/user-attachments/assets/ff9b464f-3d83-4127-ac43-d8b2f5e28b4e

smitjp150622:
> @smitjp150622 It's still not like the codepen: https://codepen.io/shaunrfox/pen/azzLVgp?editors=1100 It's not smooth at all.
> 
>  Screen.Recording.2025-05-28.at.2.36.56.PM.mov 
>  Screen.Recording.2025-05-28.at.2.37.52.PM.mov

@Etwigg, I have reviewed the video above and resolved the smooth animation issue. Please check on your end once it's available.

smitjp150622:
@shaunrfox, I added some transform and transition properties in the recipe file. I also checked our token file, and currently, there are no tokens defined for either of those properties. Should we consider adding transform and transition tokens to the token file so they can be used consistently in components like the switch recipe? Since we are already using tokens for other animation-related styles, it might be better to standardize these as well.

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,39 @@
...
+        ...props
+    }: SwitchBoxProps,
+) => {
+    const [checked, setChecked] = React.useState(false)
+    const { container, input, indicator} = switchbox({});
+    return(
+        <Label className={container} 
+        data-checked={checked ? true : undefined}
+        data-disabled={props.disabled}
+        {...(error && {'data-error' : true})}
```
Etwigg: Handle error the same way as described in the text area PR

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,39 @@
...
+        error,
+        ...props
+    }: SwitchBoxProps,
+) => {
+    const [checked, setChecked] = React.useState(false)
+    const { container, input, indicator} = switchbox({});
+    return(
+        <Label className={container} 
+        data-checked={checked ? true : undefined}
+        data-disabled={props.disabled}
```
Etwigg: I think we want to handle these the same way as error. Panda _checked will only check for the existence of data-checked, not the value if I'm not mistaken.

Etwigg (changes requested):

Etwigg:
Animation looks good!

src/components/Switch/Switch.tsx
smitjp150622: @Etwigg  Done this changes.

src/components/Switch/Switch.tsx
smitjp150622: Yes @Etwigg, I’ve updated the implementation to properly use data-checked in the relevant components/recipes.

src/components/Switch/Switch.tsx
smitjp150622: For the error fix, I’m waiting for @shaunrfox response. Once he replies, I’ll proceed with the implementation.

panda.config.ts
```diff
@@ -119,6 +120,8 @@ export default defineConfig({
         pre: preRecipe,
         spinner: spinnerRecipe,
         divider: dividerRecipe,
+        switchbox: switchRecipe,
```
Etwigg:
```suggestion
        switch: switchRecipe,
```

panda.config.ts
```diff
@@ -119,6 +120,8 @@ export default defineConfig({
...
+        //Panda might not like this since they already have a box
```
Etwigg:
```suggestion
```

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,42 @@
+import { Box, type BoxProps } from '../Box';
+import { switchbox, type SwitchboxVariantProps } from '@styled-system/recipes';
```
Etwigg:
```suggestion
import { switch, type SwitchVariantProps } from '@styled-system/recipes';
```

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { Icon } from '../Icon';
+import { Label } from '../Label';
+import React from 'react';
+
+export type SwitchBoxProps = Omit<BoxProps, keyof SwitchboxVariantProps> &
+  SwitchboxVariantProps & {
+    error?: boolean;
+    disabled?: boolean;
+  };
```
Etwigg:
```suggestion
export type SwitchProps = Omit<BoxProps, keyof SwitchVariantProps> &
  SwitchVariantProps & {
    error?: boolean;
  };
```

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const Switch: React.FC<SwitchBoxProps> = ({
+  error,
+  disabled,
+  checked,
+  ...props
+}: SwitchBoxProps) => {
```
Etwigg:
```suggestion
export const Switch: React.FC<SwitchProps> = ({
  error,
  checked,
  ...props
}: SwitchProps) => {
```

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <Box
+        as="span"
+        class={background}
+        name={'switch-bg'}
+      />
+      <Icon name={'circle'} className={indicator} />
+      <Icon name={'circle-check'} className={indicator} />
+    </Label>
+  );
+};
```
Etwigg:
```suggestion
  const { container, input, indicator, background } = switch({});
  return (
    <Label
      className={container}
    >
      <Box
        as="input"
        type="checkbox"
        className={input}
        data-checked={checked}
        {...error && { 'data-error': true }}
        {...props}
      />
      <Box
        as="span"
        class={background}
        name={'switch-bg'}
      />
      <Icon name={'circle'} className={indicator} />
      <Icon name={'circle-check'} className={indicator} />
    </Label>
  );
};
```

src/recipes/switch.ts
```diff
@@ -0,0 +1,118 @@
...
+    },
+  },
+};
+
+export const switchRecipe = defineSlotRecipe({
+  className: 'switchbox',
+  jsx: ['switchbox'],
+  slots: ['container', 'input', 'indicator', 'background'],
+  base: switchBase,
+});
```
Etwigg:
```suggestion
export const switchRecipe = defineSlotRecipe({
  className: 'switch',
  jsx: ['switch'],
  slots: ['container', 'input', 'indicator', 'background'],
  base: switchBase,
});
```

Etwigg (changes requested):
Change Switchbox to just Switch.
Couple bits we don't need.

Etwigg (changes requested):
Issues with error state css also.
Light mode + error looks like the fill isn't centered when set to checked:
image
I'm looking at the red border and the black background/fill.

Dark mode + error, the red border completely goes away when you check it.
Unchecked:
image

Checked:
image

panda.config.ts
smitjp150622: @Etwigg I previously tried using switchbox as the main component for the switch, but encountered a JavaScript error since switch is a reserved keyword and can't be used as a variable name. That’s why I kept the component name as Switchbox.

If possible, I’d suggest renaming the component to Toggle instead. It avoids the reserved keyword issue and also aligns better with common design terminology—most design systems refer to switches as toggles, which may improve clarity and usability.

Let me know your thoughts on this!

Etwigg:
@smitjp150622 
You're right! I didn't consider that. Let's name everything 'Toggle' then. I think that should work.

Etwigg:
@smitjp150622 
Conflicts!

Etwigg (approved):

PR #87, opened 2025-06-13
Feat/working package

Etwigg:
Got the repo packaged up and stuff.
Got it working on the app and with the website!


Etwigg:
ignore how many commits I made while doing this 👀

Etwigg:
The SVGs aren't going to work the way we have them set up. We need to use Vite's tool for making a sprite sheet instead of rolling our own.

.gitignore
```diff
@@ -9,8 +9,12 @@ lerna-debug.log*
 
 node_modules
 dist
+package-lock.json
```
Yash-CETEC, line 12: We will need to commit package-lock.json from now onwards, which will in future help us to maintain consistency across the local team for development. So don't add it in gitignore.

panda-preset.ts
```diff
@@ -0,0 +1,177 @@
+import {
```
Yash-CETEC, line 1: Nice 🚀 , it will help maintain consistency across the projects where this will be used. Can you add the `prefix` property in this as well, and see if what other properties are mentioned in the config file can be added here as well? Also, rather than duplicating these properties in the config file add this file to `presets` inside config file and remove the duplicate config from there.

Yash-CETEC: Especially, the prefix config should be there so that both @smitjp150622 and I can modify the existing config file with preset value and which can maintain consistancy across projects as well.

Etwigg: That's the hope. Unfortunately setting up the preset was not as straightforward as I hoped. It follows a similar, but different format than the normal panda config, so we'll need to figure that out at some point so we can bundle it with the package and have setup be easier.

Etwigg:
I think we can go ahead and merge this.
I've been making the package out of this branch for a minute.

atkinsdavid (approved):

Issue #47, opened 2025-05-16
Toggle Input


Etwigg:
Combine Toggle + Input

https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1390-4813&m=dev&t=zorehhi4oWHDjOBw-1

PR #88, opened 2025-06-19
Toggle Input complete

smitjp150622:
issue #47


src/components/ToggleInput/ToggleInput.tsx
```diff
@@ -0,0 +1,32 @@
+import { FC, ReactNode } from 'react';
+import { Box, BoxProps } from '../Box';
+import {
+  toggleInput,
+  type ToggleInputVariantProps,
+} from '@styled-system/recipes';
+import { splitProps } from '~/utils/splitProps';
+import { Label } from '../Label';
+import { cx } from '@styled-system/css';
+import { Toggle } from '../Toggle/Toggle';
```
Etwigg:
```suggestion
import { Toggle } from '../Toggle';
```

src/components/ToggleInput/ToggleInput.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const ToggleInput: FC<ToggleInputProps> = ({
+  children,
+  error,
+  ...props
+}: ToggleInputProps) => {
+  const [className, otherProps] = splitProps(props);
+  return (
+    <Label className={cx(toggleInput({}), className)} {...otherProps}>
+      <Toggle {...(error && { 'data-error': true })} {...props} />
+      {children && <Box as="div">{children}</Box>}
```
Etwigg: I don't think children need to be wrapped in Box, right? Any component we pass in here is already wrapped in box. Is it just so we have a div to put it in? If so, I think we can just put a <div> in there, right?

Etwigg (changes requested):
When you put several in the same stack there's some weird overlap happening: 
image

Etwigg:
This weird thing is happening where it's highlighting the text when I click the toggle. Can you see if there's a way to fix that?

https://github.com/user-attachments/assets/6b6f1536-050a-4d32-ab82-ddd5abd33b68

Etwigg (approved):

PR #89, opened 2025-06-25
storybook refactor

atkinsdavid:
Beginning changes to standardize Storybook formatting


.storybook/cetecTheme.tsx
```diff
@@ -0,0 +1,42 @@
+import { create } from '@storybook/theming';
+
+// TODO: Static Import Logo?
+// TODO: Implement Tokens or use DS colors
+
+export default create({
+  brandTitle: 'Cetec ERP Design System',
+  brandImage: 'https://cetecerp.com/assets/img/cetec-logo.png',
```
Etwigg: Should this be pointing to a static asset? In public or something?
It should also be an svg.
Also, I don't think that's the right logo.
And I'm not seeing where it's supposed to show on the storybook.

.storybook/preview.tsx
```diff
@@ -1,21 +1,31 @@
...
+  initialGlobals: {
+    backgrounds: { value: 'light' },
+  },
   parameters: {
+    backgrounds: {
+      options: {
+        light: { name: 'Light', value: '#F9F8F6' },
+        dark: { name: 'Dark', value: '#2E2E2E' },
+      },
+    },
```
Etwigg: I don't think this is working.

panda.config.ts
```diff
@@ -75,7 +75,11 @@ export default defineConfig({
   jsxStyleProps: 'all',
   jsxFactory: 'styled',
   watch: true,
-  include: ['./src/**/*.{js,jsx,ts,tsx}', './pages/**/*.{js,jsx,ts,tsx}'],
+  include: [
+    './src/**/*.{js,jsx,ts,tsx}',
+    './pages/**/*.{js,jsx,ts,tsx}',
+    './src/components/*/*.stories.@(js|jsx|mjs|ts|tsx)'
```
Etwigg, line 82: Different file extension syntax? Should we change the other two to this format?

.storybook/preview.tsx
Etwigg: It's working on the 'Default' pages for stories, but not on the documentation pages.

.storybook/main.ts
```diff
@@ -3,16 +3,23 @@ import type { StorybookConfig } from '@storybook/react-vite';
...
   framework: {
     name: '@storybook/react-vite',
     options: {},
   },
+  features: {
+    backgroundsStoryGlobals: true,
+  },
+  docs: {
+    defaultName: 'Documentation',
+  },
```
Etwigg, line 24: I think this is naming every documentation page 'Documentation'.
We'll probably want it to be named something unique for search purposes? I imagine we'll define that in the story itself.

image
image

Etwigg (dismissed):
Sweet. Looks like it works. I think this is a great start. We'll want to fine-tune the css and layout of everything, but as far as getting it up and running is concerned, I think we're there.

Etwigg (dismissed):
yolo

shaunrfox (approved):
This looks good for MVP. I made some small changes that got theming working.

Etwigg (commented):
@shaunrfox 
The theme seems to only be working for the components and text on component pages.
LIght:
image

Dark:
image


It also has no effect on the intro or tokens page.

If we're just wanting to move forward with a running storybook then we can merge and make a clean up pass more intentionally later.

Etwigg (approved):
Still works, so merge if you want.

Issue #64, opened 2025-05-16
Tooltip


Etwigg:
https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1614-6574&m=dev&t=zorehhi4oWHDjOBw-1

shaunrfox:
This is ready now.
https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1614-6574&p=f&t=A1u6uoMdcqEF8wBz-11

PR #90, opened 2025-06-30
Tooltip Component

smitjp150622:
issue #64


src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const Tooltip: FC<TooltipProps> = ({
+  trigger = 'onHover',
+  caret = true,
+  text,
+  title,
+  children,
+  placement = 'bottom',
+  ...props
+}) => {
+  const { wrapper, tooltipContent } = tooltip({ placement });
```
Etwigg: This was part of my pushed change

src/recipes/tooltip.ts
```diff
@@ -0,0 +1,161 @@
...
+          borderRightColor: { base: 'slate.90', _dark: 'slate.0' },
+        },
+      },
+    },
+  },
+};
+
+export const tooltipRecipe = defineSlotRecipe({
+  className: 'tooltip',
+  jsx: ['Tooltip'],
```
Etwigg, line 203: This too

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const handleClick = () => {
+    if (trigger === 'onClick') setShow((prev) => !prev);
+  };
+
+  return (
+    <Box as="div" {...props} className={wrapper} ref={tooltipRef}>
+      <Box
+        onMouseEnter={trigger === 'onHover' ? handleMouseEnter : undefined}
+        onMouseLeave={trigger === 'onHover' ? handleMouseLeave : undefined}
+        onClick={trigger === 'onClick' ? handleClick : undefined}
```
Etwigg, line 229: I'm a bit confused why we're redefining these event handlers like this. You can just add normal onHover can you not?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+      >
+        {children}
+      </Box>
+
+      {show && (
+        <Box as="span" className={tooltipContent}>
+          {title && <Box as="p">{title}</Box>}
```
Etwigg: This should be body/md bold text.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+          {text && <Box as="span">{text}</Box>}
```
Etwigg: Why are we using Box here not the text component?

src/components/Tooltip/Tooltip.tsx
Etwigg: also why are the event handlers on their own box component?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <Box
+        onMouseEnter={trigger === 'onHover' ? handleMouseEnter : undefined}
+        onMouseLeave={trigger === 'onHover' ? handleMouseLeave : undefined}
+        onClick={trigger === 'onClick' ? handleClick : undefined}
+      >
+        {children}
+      </Box>
+
+      {show && (
+        <Box as="span" className={tooltipContent}>
```
Etwigg:
```suggestion
        <Box className={tooltipContent}>
```

should be div

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const handleMouseLeave = () => {
+    if (trigger === 'onHover') setShow(false);
+  };
+
+  const handleClick = () => {
+    if (trigger === 'onClick') setShow((prev) => !prev);
+  };
+
+  return (
+    <Box as="div" {...props} className={wrapper} ref={tooltipRef}>
```
Etwigg:
```suggestion
    <Box {...props} className={wrapper} ref={tooltipRef}>
```

box defaults to div

smitjp150622: Okay I got it.

src/components/Tooltip/Tooltip.tsx
smitjp150622: When using a normal onHover, we typically add a class and manage visibility through CSS. similarly, for onClick, the tooltip should behave the same—becoming visible on interaction. that’s why I thought it would be better to handle both onHover and onClick using Box event handlers, so the behavior remains consistent and manageable at the component level.

src/components/Tooltip/Tooltip.tsx
smitjp150622: Ok I will Change it.

src/components/Tooltip/Tooltip.tsx
smitjp150622: I think there may be cases where we need to render raw HTML content inside the tooltip’s content section. In such situations, we can use:

Example:- `dangerouslySetInnerHTML={{ __html: <p> First, Second </p>}}`

If there’s no need for HTML content, i can fall back to using the Text component instead.

src/components/Tooltip/Tooltip.tsx
smitjp150622: Also, as per @shaunrfox's reference to the MUI Tooltip, i checked their implementation as well, it behaves the same way, using mouse events to handle visibility.

src/components/Tooltip/Tooltip.tsx
Etwigg: Fair point, but maybe that's something we should be handling with the Text component as well?
TextArea is built off of text and will need to handle HTML.

Etwigg (changes requested):
Left-End looks like it's starting at the top: 
image

vs. 
image

I think you just have start and end backwards for all of these.

Margin is wrong for tooltips with the caret option. With caret is 12 without is 8, it looks like all of these are 8 margin.

Getting some weird flickering on hover:

https://github.com/user-attachments/assets/fce73389-8b37-4a09-9d6a-ce32a0fc1fcf

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,276 @@
+import { Box, type BoxProps } from '../Box';
+import { tooltip, type TooltipVariantProps } from '@styled-system/recipes';
+import { FC, ReactNode, useEffect, useRef, useState } from 'react';
```
Etwigg: This is a nit-pick, but for style sake can we not import FC? We use React.FC everywhere else.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,276 @@
...
+  'left',
+  'top-start',
+  'top-end',
+  'right-start',
+  'right-end',
+  'bottom-end',
+  'bottom-start',
+  'left-start',
+  'left-end',
+];
```
Etwigg: Why do we need to do this?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,276 @@
...
+      'top-start': ['top-end', 'top', 'bottom-start', 'left-start', 'right-start'],
+      'top-end': ['top-start', 'top', 'bottom-end', 'left-end', 'right-end'],
+      'top': ['top-start', 'top-end', 'bottom', 'left', 'right'],
+      'left-start': ['left-end', 'left', 'right-start', 'top-start', 'bottom-start'],
+      'left-end': ['left-start', 'left', 'right-end', 'top-end', 'bottom-end'],
+      'left': ['left-start', 'left-end', 'right', 'top', 'bottom'],
+      'right-start': ['right-end', 'right', 'left-start', 'top-start', 'bottom-start'],
+      'right-end': ['right-start', 'right', 'left-end', 'top-end', 'bottom-end'],
+      'right': ['right-start', 'right-end', 'left', 'top', 'bottom'],
+    };
```
Etwigg: Was there not a programatic way to handle this?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,276 @@
...
+        if (fits) {
+          setCurrentPlacement(positioning);
+          return;
+        }
+      }
+      setCurrentPlacement(resolvedPlacement);
+    };
+    
+
+  type Placement = 'top' | 'bottom' | 'left' | 'right' | 'top-start' | 'bottom-start' | 'left-start' | 'right-start' | 'top-end' | 'bottom-end' | 'left-end' | 'right-end'
```
Etwigg: Why redefine this type? Don't you already have this with ```TooltipVariantProps['placement']```

src/components/Tooltip/Tooltip.tsx
smitjp150622: Remove, we not need to this.

Etwigg:
@smitjp150622 
Is this good?

src/components/Tooltip/Tooltip.tsx
smitjp150622: Resolve this issue.

smitjp150622:
> @smitjp150622 Is this good?
yes, also i resolved the conflict also.

smitjp150622:
> Left-End looks like it's starting at the top: image
> 
> vs. image
> 
> I think you just have start and end backwards for all of these.
> 
> Margin is wrong for tooltips with the caret option. With caret is 12 without is 8, it looks like all of these are 8 margin.
> 
> Getting some weird flickering on hover:
> 
>  Screen.Recording.2025-07-09.at.6.55.15.PM.mov

Flickering issue resolved, also the with caret and without caret margin issue resolved. and for the start and end point i flowed the example as per shaun provide in figma. https://codepen.io/shaunrfox/pen/XJbpRPx?editors=1100

Etwigg:
@smitjp150622 
If everything is done and it's ready for review, you can mark it ready for review and take it out of draft.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+export type TooltipProps = Omit<BoxProps, keyof TooltipVariantProps> &
+  TooltipVariantProps & {
+    text?: string;
+    title?: string;
+    caret?: boolean;
+    placement?: Placement;
+    children?: string | ReactNode;
+    trigger?: 'onHover' | 'onClick';
+  };
```
Etwigg: placement -> position

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,259 @@
...
+import { Text } from '../Text';
+
+export type Placement =
+  | 'top' | 'bottom' | 'left' | 'right'
+  | 'top-start' | 'bottom-start' | 'left-start' | 'right-start'
+  | 'top-end' | 'bottom-end' | 'left-end' | 'right-end';
+
+export type TooltipProps = Omit<BoxProps, keyof TooltipVariantProps> &
+  TooltipVariantProps & {
+    text?: string;
```
Etwigg: text shouldn't be optional

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,259 @@
...
+    title?: string;
+    caret?: boolean;
+    placement?: Placement;
+    children?: string | ReactNode;
```
Etwigg: I don't think children should ever be string, right?

Etwigg (changes requested):
Seeing a seam sometimes between the carat and the box?

Image

src/components/Tooltip/Tooltip.tsx
smitjp150622: resolved.

src/components/Tooltip/Tooltip.tsx
smitjp150622: resolved.

src/components/Tooltip/Tooltip.tsx
smitjp150622: Done

smitjp150622:
@Etwigg all changes are done please review this.

Etwigg (dismissed):

Etwigg:
@abderahmentrabelsi 
If you could take a look at this, I'd appreciate it! Feel free to merge once you approve.

src/components/Toggle/Toggle.tsx
```diff
@@ -32,7 +32,7 @@ export const Toggle: React.FC<ToggleProps> = ({
       />
```
abderahmentrabelsi, line 32: can we render data-checked only when checked is true to avoid [data-checked] selectors matching the unchecked state?

abderahmentrabelsi (dismissed):
suggestion only , LGTM

src/components/Toggle/Toggle.tsx
smitjp150622: Done.

smitjp150622: Done

abderahmentrabelsi (approved):

Etwigg (approved):

PR #92, opened 2025-08-01
Playroom implementation

atkinsdavid:
Initial work toward implementing playroom. 

Current status is that Playroom will load, but does not seem to be able to access the components from our library. 

I believe the fix will be somewhere between the `webpackConfig` in `playroom.config.cjs` and the Frame component in `playroom/frameComponent.tsx`. Playroom needs a wrapper component in order to load the custom components into the UI, and because playroom isn't currently packaged where it can be deployed with Vite you have to make sure that you've got the right interpreters to parse the TS files in a way that Playroom can render them.


atkinsdavid:
Also, sorry about all the commented out bits, there's been a lot of trial and error trying to get things working.

Etwigg:
@atkinsdavid 
78k additions?
webpack?
I think you got the static directory committed here.

shaunrfox:
Playroom now works locally and builds to ./dist/playroom. There's more optimization we could do and some small bugs, but it's ready as a v1. This also revealed some issues in our tokens and other styles, so you'll see some fixes too.

Etwigg:
@shaunrfox @atkinsdavid 
Can we clean up the... 144k additions? No way that's correct, right?

shaunrfox:
I'll take a look



Shaun Fox

***@***.***

512.554.2140


3232 E Cesar Chavez, Suite 2.150
Austin, TX 78702

cetecerp.com

Issue #51, opened 2025-05-16
FormField


Etwigg:
Combination of inputs for use together.

https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1396-7385&m=dev&t=zorehhi4oWHDjOBw-1

shaunrfox:
This is ready now.
https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1396-7385&p=f&t=A1u6uoMdcqEF8wBz-11

PR #94, opened 2025-08-08
feat: formfield component

smitjp150622:
issue #51

### DIFF
--- a/src/components/CheckBox/CheckBox.tsx
+++ b/src/components/CheckBox/CheckBox.tsx
@@ -18,6 +18,7 @@ export const CheckBox: React.FC<CheckBoxProps> = ({
   error,
   id,
   name,
+  checked,
   ...props
 }) => {
   const { container, input, indicator } = checkbox({});
@@ -35,6 +36,7 @@ export const CheckBox: React.FC<CheckBoxProps> = ({
         id={id}
         aria-label={name}
         {...props}
+        {...(checked ? { 'data-checked': true } : {})}
         {...(indeterminate && { 'data-indeterminate': true })}
         {...(error && { 'data-error': true })}
       />
--- a/src/components/CheckBox/CheckBox.tsx
+++ b/src/components/CheckBox/CheckBox.tsx
@@ -36,7 +36,6 @@ export const CheckBox: React.FC<CheckBoxProps> = ({
         id={id}
         aria-label={name}
         {...props}
-        {...(checked ? { 'data-checked': true } : {})}
         {...(indeterminate && { 'data-indeterminate': true })}
         {...(error && { 'data-error': true })}
       />
--- a/src/components/FormField/FormField.tsx
+++ b/src/components/FormField/FormField.tsx
@@ -22,7 +22,7 @@ export type FormFieldProps = Omit<BoxProps, keyof FormFieldVariantProps> &
   };
 
 export const FormField: React.FC<FormFieldProps> = ({
-  layout,
+  layout = "default",
   label,
   helpText,
   required,
@@ -54,19 +54,21 @@ export const FormField: React.FC<FormFieldProps> = ({
   return (
     <Box className={formFeildContainer} {...props} disabled={disabled}>
       <Box className={labelWrapper}>
-        <Text textStyle="body-md" className={headLabel}>
-          {label}{' '}
-          {required && (
-            <Text as="span" color="red.50">
-              *
-            </Text>
-          )}
-          {tooltip && (
-            <Tooltip title={tooltipTitle} text={`${tooltipDescription}`} caret={tooltipCaret}>
-              <Icon name="info" fill='slate.50'/>
-            </Tooltip>
-          )}
-        </Text>
+        <Box className={headLabel}>
+          <Text textStyle="body-md">
+            {label}{' '}
+          </Text>
+            {required && (
+              <Text as="span" color="red.50">
+                *
+              </Text>
+            )}
+            {tooltip && (
+              <Tooltip title={tooltipTitle} text={`${tooltipDescription}`} caret={tooltipCaret}>
+                <Icon name="info" fill='slate.50'/>
+              </Tooltip>
+            )}
+        </Box>
         {layout === "default" && helpText && (
           <Text as="span" textStyle="body-sm">
             {helpText}
--- a/src/components/FormField/FormField.tsx
+++ b/src/components/FormField/FormField.tsx
@@ -1,0 +1,66 @@
+import { formField, type FormFieldVariantProps } from '@styled-system/recipes';
+import { Box, type BoxProps } from '../Box';
+import { Text } from '../Text';
+
+export type FormFieldProps = Omit<BoxProps, keyof FormFieldVariantProps> &
+  FormFieldVariantProps & {
+    layout?: 'default' | 'inline',
+    label?: string;
+    helpText?: string;
+    required?: boolean;
+    infoTooltip?: string;
+    error?: boolean;
+    errorText?: string;
+    disabled?: boolean;
+    children?: React.ReactNode;
+  };
+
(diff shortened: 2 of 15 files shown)


Etwigg:
@abderahmentrabelsi @Yash-CETEC 
Can we get some reviews on this?
Thanks

Etwigg (changes requested):
Please provide more examples of other inputs being used as well as props that we're using for the form field itself.
i.e. Inline vs default
You have one example of inline help text, but the label isn't inline:
image

smitjp150622:
> Please provide more examples of other inputs being used as well as props that we're using for the form field itself. i.e. Inline vs default You have one example of inline help text, but the label isn't inline: image

Resolved variant issue, and add more example.

Etwigg (changes requested):
The radio acts a bit strange. You can select all of the radio options and clicking them again doesn't deselect.

https://github.com/user-attachments/assets/b8d1c04f-3b53-47d1-a93b-8d1cfffdfe9c

smitjp150622:
> The radio acts a bit strange. You can select all of the radio options and clicking them again doesn't deselect.
> 
>  Screen.Recording.2025-09-19.at.4.15.03.PM.mov

issue resolved.

Etwigg (dismissed):

Etwigg:
@abderahmentrabelsi 
Can I get a review here please?
Thanks!

Etwigg (changes requested):
inline + error state doesn't work: 
image

image

Info tooltip doesn't work.
Form field 'error' does not propagate to children.
'disabled' prop doesn't do anything. You're missing the whole disabled state and not passing it to children.

Etwigg:
help text and error text don't work when a radio component is the child.

Please test this more thoroughly before handing it back.

Etwigg (dismissed):
I think we're good now.

Etwigg:
@abderahmentrabelsi can you take a pass and approve/merge please

src/components/Toggle/Toggle.tsx
```diff
@@ -36,7 +36,7 @@ export const Toggle: React.FC<ToggleProps> = ({
         name={'toggle-bg'}
       />
       <Icon name={'circle'} className={indicator} />
-      <Icon name={'circle-check'} className={indicator} />
+      <Icon name={'circle-check'} className={indicator} fill={{ _dark: 'slate.90'}}/>
```
abderahmentrabelsi: type!!

smitjp150622: Yes, I’ve already made this change in the Menu component PR, so once that’s merged, it won’t be needed in the Toggle file.

src/components/FormField/FormField.tsx
```diff
@@ -0,0 +1,93 @@
...
+      const c = child as React.ReactElement<any>;
+      return React.cloneElement(c, {
+        error: error ?? c.props.error,
+        // disabled: disabled ?? c.props.disabled,
+      });
+    }
+    return child;
+  });
+  return (
+    <Box className={formFeildContainer} {...props} disabled={disabled}>
```
abderahmentrabelsi: can we merge consumer className with the recipe class (e.g., cx(formFieldClass, props.className)) instead of overwriting it by spreading {...props} after className?

src/components/FormField/FormField.tsx
```diff
@@ -0,0 +1,93 @@
...
+        </Box>
+        {layout === "default" && helpText && (
+          <Text as="span" textStyle="body-sm">
+            {helpText}
+          </Text>
+        )}
+      </Box>
+      <Box className={contentWrapper}>
+        {enhancedChildren}
+        {layout == 'inline' && (
```
abderahmentrabelsi: loose equality check , switch to === ?

src/recipes/formFieldRecipe.ts
```diff
@@ -0,0 +1,57 @@
+import { defineSlotRecipe } from '@pandacss/dev';
+
+const formFieldBase = {
+  formFeildContainer:{
```
abderahmentrabelsi: typo maybe to avoid misspelling?

src/recipes/formFieldRecipe.ts
```diff
@@ -0,0 +1,57 @@
...
+    _disabled:{
```
abderahmentrabelsi: i think using :disabled style on a div wont aply  can we switch to a data-disabled selector (and set data-disabled on the container)?

src/components/FormField/FormField.tsx
smitjp150622: @abderahmentrabelsi I used {...props} not just for the class name — it’s for handling any box related props when needed. that’s why i had include {...props}.

src/recipes/formFieldRecipe.ts
smitjp150622: data-disabled already apply.
Screenshot 2025-10-14 at 3 21 57 PM

src/components/CheckBox/CheckBox.tsx
```diff
@@ -13,17 +13,21 @@ export type CheckBoxProps = Omit<BoxProps, keyof CheckboxVariantProps> &
...
+  checked,
   ...props
 }) => {
   const { container, input, indicator } = checkbox({});
 
   return (
-    <Label className={container}>
+    <Label className={container} color={error
+      ? 'red.50'
+      : { base: 'slate.90', _dark: 'slate.0' }}>
```
Etwigg: I don't think I like handling the color here. We should be able to use a condition in the recipe to check for error and the default recipe should handle the rest.
TLDR: We should just pass error as a prop to Label and the Label recipe should handle it.

src/components/CheckBox/CheckBox.tsx
```diff
@@ -13,17 +13,21 @@ export type CheckBoxProps = Omit<BoxProps, keyof CheckboxVariantProps> &
...
       <Box
         as="input"
         type="checkbox"
         className={input}
         {...props}
+        {...(checked ? { 'data-checked': true } : {})}
```
Etwigg: I don't think we need to explicitly list this here, right? The box is a checkbox, it has 'checked' by default.

src/components/Radio/Radio.tsx
```diff
@@ -12,7 +12,9 @@ export type RadioProps = Omit<BoxProps, keyof RadioVariantProps> &
 export const Radio: React.FC<RadioProps> = ({ error, ...props }) => {
   const { container, input, indicator } = radio({});
   return (
-    <Label className={container}>
+    <Label className={container} color={error
+      ? 'red.50'
+      : { base: 'slate.90', _dark: 'slate.0' }}>
```
Etwigg: same as other label

Etwigg (changes requested):
One more change and I think this is good to merge.

src/components/CheckBox/CheckBox.tsx
smitjp150622: In the Icon component, Shaun already added `fill="current"`, which makes the icon inherit the color from its parent. because of this, i can't directly change the icon color using the recipe. i also checked how the color is applied in the Icon component and confirmed that it inherits the parent element’s color. i have attached images for your reference.

This icon use fill props `fill='current'` in `Icon.tsx`.
Screenshot 2025-10-17 at 12 17 17 PM
Also, how to use the color in the icon `App.tsx` 
Screenshot 2025-10-17 at 12 19 16 PM

Etwigg: I was talking about the label component though. Are you saying that if you try to apply it to label it doesn't work if it's wrapping an Icon?

abderahmentrabelsi:
resolve conflicts please

smitjp150622:
> resolve conflicts please

@abderahmentrabelsi conflict resolved.

PR #95, opened 2025-08-12
feat: breadcrumbs component

smitjp150622:


Etwigg:
@Yash-CETEC @abderahmentrabelsi 
Reviews here too please!

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const Breadcurmbs: React.FC<BreadcurmbsProps> = ({
+    items,
+    ...props
+}:BreadcurmbsProps) => {
+    const [className, otherProps] = splitProps(props);
+
+    return(
+        <Text as="ul" className={cx(breadcurmbs({}), className)} {...otherProps}>
+                {items?.map((item, index)=>(
```
abderahmentrabelsi, line 22: stable key instead of index , give items a named type and/or annotate the callback ?

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,30 @@
+import { breadcurmbs, BreadcurmbsVariantProps } from "@styled-system/recipes";
```
abderahmentrabelsi: no exported member 'breadcurmbs'/'BreadcurmbsVariantProps' from @styled-system/recipes , can we align naming ?

src/components/Breadcrumbs/Breadcrumbs.tsx
smitjp150622: Done.

src/components/Breadcrumbs/Breadcrumbs.tsx
smitjp150622: Done

Etwigg (changes requested):
I believe selected should also have the link styling:

Image
vs.

Image

apart from that, looks good.

Yash-CETEC (dismissed):

abderahmentrabelsi (dismissed):
LGTM

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,30 @@
+import { breadcrumbs, BreadcrumbsVariantProps } from "@styled-system/recipes";
+import { Box, type BoxProps } from "../Box";
```
Etwigg: You're not using Box, why are we importing it?

Etwigg (commented):
Looks fine except for the Box import.

src/components/Breadcrumbs/Breadcrumbs.tsx
smitjp150622: Previously, i thought using box try to make this component, but after that i use text and i forgot to remove box.

Etwigg (approved):

Issue #59, opened 2025-05-16
Card


Etwigg:
https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1494-306&m=dev&t=zorehhi4oWHDjOBw-1

PR #96, opened 2025-08-18
Card

smitjp150622:
issue #59

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -1,0 +1,37 @@
+import { card, CardVariantProps } from "@styled-system/recipes";
+import { Box, BoxProps } from "../Box";
+import { ReactNode } from "react";
+import { splitProps } from "~/utils/splitProps";
+import { cx } from "@styled-system/css";
+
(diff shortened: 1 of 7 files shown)


src/components/Card/Card.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { splitProps } from '~/utils/splitProps';
+import { cx } from '@styled-system/css';
+
+export type CardProps = BoxProps &
+  CardVariantProps & {
+    href?: string;
+    children?: string | ReactNode;
+    grabbed?: boolean;
+    disabled?: boolean;
+  };
```
Etwigg: Missing Omit<BoxProps, keyof CardVariantProps>

Etwigg (changes requested):
Something wrong with focus state on default variant:
https://github.com/user-attachments/assets/95963448-6681-4ea6-97c7-cd8ec535f708

src/components/Card/Card.tsx
smitjp150622: Done,

Etwigg (approved):

PR #97, opened 2025-08-27
Add Card, Toggle, ToggleInput recipe add in panda-preset

smitjp150622:
I added the Card, Toggle, and ToggleInput recipes in the panda-preset because I was still facing the same issue: "The requested module 'cetec-design-system' does not provide an export named 'Card'".
So there might be a chance this update will resolve the issue.


Etwigg (approved):
ok

Issue #61, opened 2025-05-16
Menu/MenuItem


Etwigg:
Placeholder

Etwigg:
https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1494-307&m=dev&t=7r4VATrnvwyUbdun-1

PR #102, opened 2025-09-18
feat: menu component

smitjp150622:
issue: #61

### DIFF
--- a/src/components/CheckBox/CheckBox.tsx
+++ b/src/components/CheckBox/CheckBox.tsx
@@ -23,7 +23,10 @@ export const CheckBox: React.FC<CheckBoxProps> = ({
   const { container, input, indicator } = checkbox({});
 
   return (
-    <Label className={container} htmlFor={id}>
+    <Label
+      className={container}
+      color={error ? 'red.50' : { base: 'slate.90', _dark: 'slate.0' }}
+    >
       <Box
         as="input"
         type="checkbox"
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -3,7 +3,7 @@ import { Box, type BoxProps } from '../Box';
 import { useState } from 'react';
 import { Text } from '../Text';
 import { Divider } from '../Divider';
-import { Icon } from '../Icon';
+import { Icon, type IconNamesList } from '../Icon';
 import { CheckBox } from '../CheckBox';
 import { Toggle } from '../Toggle';
 import { Link } from '../Link';
@@ -63,8 +63,14 @@ export const Menu: React.FC<MenuProps> = ({
 
   const handleSelect = (id: string) => {
     if (variant === 'single-select') {
-      setSelected([id]);
-      onChange?.(id);
+      if(selected.includes(id)){
+        setSelected([]);
+        onChange?.(null);
+      }
+      else{
+        setSelected([id]);
+        onChange?.(id);
+      }
     } else {
       setSelected((prev) =>
         prev.includes(id) ? prev.filter((x) => x !== id) : [...prev, id],
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -16,7 +16,6 @@ export type MenuProps = Omit<BoxProps, keyof MenuVariantProps> &
       divider?: boolean;
       spacer?: boolean;
       link?: boolean;
-      icon?: boolean;
       items: {
         id: string;
         label: string;
@@ -161,8 +160,11 @@ export const Menu: React.FC<MenuProps> = ({
                     role="button"
                     aria-pressed={isSelected}
                   >
-                    {section?.icon && (
-                      <Box className={iconSection}>
+                    {(iconPlacement || item?.iconName) && (
+                      <Box
+                        className={iconSection}
+                        color={{ base: 'slate.90', _dark: 'slate.0' }}
+                      >
                         {item?.iconName && (
                           <Icon name={`${item?.iconName as IconNamesList}`} />
                         )}
@@ -209,7 +211,10 @@ export const Menu: React.FC<MenuProps> = ({
                       </Link>
                     )}
                     {hasChildren && (
-                      <Box className={multiLevelIcon}>
+                      <Box
+                        className={multiLevelIcon}
+                        color={{ base: 'slate.90', _dark: 'slate.0' }}
+                      >
                         <Icon name="caret-right" />
                       </Box>
                     )}
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -3,7 +3,7 @@ import { Box, type BoxProps } from '../Box';
 import { useState } from 'react';
 import { Text } from '../Text';
 import { Divider } from '../Divider';
-import { Icon, type IconNamesList } from '../Icon';
+import { Icon } from '../Icon';
 import { CheckBox } from '../CheckBox';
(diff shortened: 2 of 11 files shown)


Etwigg (changes requested):
Getting a ton of errors on build.
image
image
image

etc....

Etwigg (changes requested):
Checkbox doesn't work
No divider in dark mode:
image

Etwigg (changes requested):
Clicking the box on the checkbox menu doesn't check the box. You have to click the menu item itself.
The last item here is a different style: 
image

Clicking on a menu item selects it, but clicking it again doesn't deselect it.

I don't think the underline on hover is correct here (maybe I'm missing it in the figma?): 
image

Etwigg (changes requested):
checkbox still wrong
css for tab select is wrong
description example still has different color text?
image


Menuitem background color is wrong.

Please do some A/B testing between the figma and what you have. Once you've done that let me know and I'll continue reviewing.

smitjp150622:
@Etwigg  Description and title both are same as figma.
Screenshot 2025-09-30 at 11 59 41 AM
Screenshot 2025-09-30 at 11 57 48 AM
Title
Screenshot 2025-09-30 at 11 58 09 AM
Screenshot 2025-09-30 at 11 58 35 AM

Etwigg:
@smitjp150622 
Description is absolutely not the same... come on.
Screenshot 2025-10-06 at 9 42 35 PM
The last item here is a different color than the rest. It even shows in your light mode screenshot. There's no different color last element on figma.
image
The selected state is even different between items:
https://github.com/user-attachments/assets/b6637d3a-a2e0-4bb9-a0b8-29dfb87cea33

smitjp150622:
The last item is basically a disabled example, so it should have reduced opacity to make it look different in color.
Screenshot 2025-10-07 at 10 08 19 AM

Screenshot 2025-10-07 at 10 10 17 AM

Etwigg (changes requested):
Setting the iconplacement prop makes menu entries without icons behave oddly: 
image

The padding looks off here:
current implementation: 
image


figma:
image


But the css looks right... maybe I'm just seeing things?

I am seeing a lot of duplicated css also.

smitjp150622:
@Etwigg Is it possible to have an icon in a single menu list item? because based on the Figma design, it seems that when we use icons in the menu, all items are expected to have icons.

smitjp150622:
@Etwigg right now, i have added a prop called icon. when **icon={true}** is passed to the menu component, we can use either **multiple icons** or a **single icon** in the menu list.
Screenshot 2025-10-10 at 11 43 03 AM
Screenshot 2025-10-10 at 11 43 35 AM

src/components/Toggle/Toggle.tsx
```diff
@@ -36,7 +36,7 @@ export const Toggle: React.FC<ToggleProps> = ({
         name={'toggle-bg'}
       />
       <Icon name={'circle'} className={indicator} />
-      <Icon name={'circle-check'} className={indicator} />
+      <Icon name={'circle-check'} className={indicator} fill={{ _dark: 'slate.90'}}/>
```
abderahmentrabelsi: the fill prop doesn’t accept an object value like { _dark: "slate.90" }

src/recipes/menu.ts
```diff
@@ -0,0 +1,149 @@
...
+    bg: { base: 'slate.0', _dark: 'slate.80' },
+    py: { base: '12', md: '4' },
+    boxShadow: 'medium',
+    borderRadius: {
+      base: '8',
+      md: '4',
+    },
+    borderBottomLeftRadius: { base: '0', md: '4' },
+    borderBottomRightRadius: { base: '0', md: '4' },
+    "& ~ [name='caret-left'], & ~ [name='caret-right'], & ~ svg": {
```
abderahmentrabelsi: i think & ~ svg targets all following SVGs

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,228 @@
...
+          />
+          {current?.parentLabel || 'Back'}
+        </Text>
+      )}
+
+      <Box
+        data-anim={isChildren.length > 1 ? 'slide-left' : undefined}
+        className={wrapperInner}
+      >
+        {current?.menu?.map((section) => (
```
abderahmentrabelsi, line 117: parameter 'section' implicitly has an 'any' type , can we annotate it maybe?
same for parameter item line 129

src/recipes/menu.ts
smitjp150622: Done.

src/components/Toggle/Toggle.tsx
smitjp150622: Done.

src/components/Menu/Menu.tsx
smitjp150622: I think there’s no need to add the any type because I have already declared it in the main MenuProps.

Etwigg:
> @Etwigg right now, i have added a prop called icon. when **icon={true}** is passed to the menu component, we can use either **multiple icons** or a **single icon** in the menu list. Screenshot 2025-10-10 at 11 43 03 AM Screenshot 2025-10-10 at 11 43 35 AM

I don't really like having to tell the menu component that we're going to be using icons within it. The example I gave seemed to break the css of the rest of the menuitems. I feel like there might be some css at the menu level that belongs on the menuitem level and that would prevent the issue I pointed out from happening. 
Is there not another solution to the issue?

smitjp150622:
@Etwigg, I removed the props and handled it with CSS. I also managed it using the current props.

src/components/CheckBox/CheckBox.tsx
```diff
@@ -18,7 +18,7 @@ export const CheckBox: React.FC<CheckBoxProps> = ({
   const { container, input, indicator } = checkbox({});
 
   return (
-    <Label className={container}>
+    <Label className={container} color={error ? 'red.50' : {base: 'slate.90', _dark: 'slate.0'}}>
```
Etwigg: Mentioned this in another PR. We can handle error by passing the prop to label and using the recipe, right?

src/recipes/toggle.ts
```diff
@@ -102,7 +103,7 @@ const toggleBase = {
     transitionDuration: '200ms',
     transitionTimingFunction: 'ease-in-out',
     "&:is([name='circle'])": {
-      fill: { base: 'slate.20', _dark: 'slate.40' },
+      fill: { base: 'slate.20 !important', _dark: 'slate.40 !important' },
```
Etwigg, line 106: Why are we using important here?

Etwigg (changes requested):

src/recipes/toggle.ts
smitjp150622: Because the Icon component uses `fill: 'current'`, the icon inherits the color from its parent element. That’s why i had to use !important, or alternatively, i can set the color directly on the parent element like this: `color={error ? 'red.50' : { base: 'slate.90', _dark: 'slate.0' }}`.

src/components/CheckBox/CheckBox.tsx
smitjp150622: Please check below comment link.
https://github.com/Cetec-ERP/Cetec-Design-System/pull/94#discussion_r2438597822

Etwigg:
@smitjp150622 
Is this ready to review again?

smitjp150622:
yes, this is ready now.

Issue #53, opened 2025-05-16
Tag


Etwigg:
https://www.figma.com/design/ZIKwQGwbWwxUQ0MEodl5vi/Cetec-Components?node-id=1484-626&m=dev&t=zorehhi4oWHDjOBw-1

PR #112, opened 2025-10-08
feat: tag component

smitjp150622:
issue:- #53

### DIFF
--- a/src/components/Tag/Tag.tsx
+++ b/src/components/Tag/Tag.tsx
@@ -1,32 +1,35 @@
-import { tag, type TagVariantProps } from "@styled-system/recipes";
-import { Box, type BoxProps } from "../Box";
-import { ReactNode } from "react";
-import { splitProps } from "~/utils/splitProps";
-import { cx } from "@styled-system/css";
-import { Icon, type IconNamesList } from "../Icon";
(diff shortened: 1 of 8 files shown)


Etwigg (changes requested):
You still need to make the contrast changes, correct? In the figma we've got a lighter color for the text on the off white tags.

smitjp150622:
@Etwigg all 17 color contrast checks are completed.

src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,32 @@
+import { tag, type TagVariantProps } from "@styled-system/recipes";
+import { Box, type BoxProps } from "../Box";
+import { ReactNode } from "react";
+import { splitProps } from "~/utils/splitProps";
+import { cx } from "@styled-system/css";
+import { Icon, type IconNamesList } from "../Icon";
+
+export type TagProps = BoxProps & TagVariantProps & {
```
abderahmentrabelsi: does TagVariantProps already include variant/hue/iconPosition/hasIcon (see tagRecipe) ?  If yes can we remove iconPosition here and type iconName: IconNamesList (so no cast at Icon render)?

smitjp150622: Done.

Etwigg (dismissed):

PR #113, opened 2025-10-16
feat: accessibility standards

smitjp150622:


src/components/CheckBox/CheckBox.tsx
```diff
@@ -2,27 +2,35 @@ import { Box, type BoxProps } from '../Box';
...
+import { AriaAttributes } from 'react';
 
 export type CheckBoxProps = Omit<BoxProps, keyof CheckboxVariantProps> &
   CheckboxVariantProps & {
+    name: string;
     indeterminate?: boolean;
     disabled?: boolean;
     error?: boolean;
-  };
+    id?: string;
```
Etwigg: I don't think we can do id. At least not like this. It would have to be some uuid probably. 
Label doesn't need htmlfor, it applies to the immediate child.

Etwigg, line 13: If we're using htmlfor then we'll require an ID be passed, right? If we don't then we're not going to pass accessibility.

Etwigg:
@smitjp150622 
Let me know when this is ready for review!

src/components/CheckBox/CheckBox.tsx
smitjp150622: No, only name is required, if we pass either name or id so pass the accessibility.

Etwigg: won't that break the htmlfor on the label if we don't pass id?

Etwigg: nvm, I guess it doesn't matter if we use the htmlfor, it should default to the implicit behavior if we don't pass in ID

Etwigg (approved):
good first pass

PR #116, opened 2025-11-11
refactor!: checkbox updates

shaunrfox:


src/components/CheckBox/CheckBox.tsx
```diff
@@ -4,20 +4,22 @@ import { checkbox, type CheckboxVariantProps } from '@styled-system/recipes';
...
-export type CheckBoxProps = Omit<BoxProps, keyof CheckboxVariantProps> &
+export type CheckboxProps = Omit<BoxProps, keyof CheckboxVariantProps> &
   CheckboxVariantProps & {
     name: string;
     indeterminate?: boolean;
     disabled?: boolean;
     error?: boolean;
-    id?: string;
+		id?: string;
+		labelledby?: string;
```
Etwigg: Formatting issues?

src/components/CheckBox/CheckBox.tsx
```diff
@@ -4,20 +4,22 @@ import { checkbox, type CheckboxVariantProps } from '@styled-system/recipes';
...
   } & AriaAttributes;
 
-export const CheckBox: React.FC<CheckBoxProps> = ({
+export const Checkbox: React.FC<CheckboxProps> = ({
   indeterminate,
   error,
   id,
-  name,
+	name,
+	labelledby,
```
Etwigg: formatting

src/components/CheckBox/CheckBox.tsx
```diff
@@ -33,10 +35,11 @@ export const CheckBox: React.FC<CheckBoxProps> = ({
         className={input}
         name={name}
         id={id}
-        aria-label={name}
-        {...props}
+				aria-label={name}
+				{...labelledby && { 'aria-labelledby': labelledby }}
```
Etwigg: formatting

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -1,35 +1,53 @@
...
         {...(error && { 'data-error': true })}
         {...(indeterminate && { 'data-indeterminate': true })}
         {...props}
       />
-      {children && <Box as="span">{children}</Box>}
+      {children && (
+        <Text as="span" id={`${id}-description`}>
+          {children}
+        </Text>
+      )}
```
Etwigg: This is a weird pattern.
We'd normally be passing the text component AS the child, right?
Doing it this way you'd just wrap a prop in the text component instead of having text wrap children. It limits us to only strings as children too.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -1,35 +1,53 @@
...
+}: CheckboxInputProps) => {
+  const [className, otherProps] = splitProps(props);
   return (
-    <Label htmlFor={id}>
-      <CheckBox
+    <Label
+      className={cx(checkboxInput(), className)}
+      {...otherProps}
+      htmlFor={id}
+    >
```
Etwigg, line 45: I'm pretty sure this is wrong.
Label has its own className that's applied in that component. It shouldn't be getting the styles from checkboxInput.

Etwigg (changes requested):

src/components/CheckboxInput/CheckboxInput.tsx
shaunrfox: Yeah, you're right. This was existing. I'll address it

shaunrfox:
Updates:
- Checkbox component - Converted to controlled-only pattern (required checked and onChange props), conditional icon rendering (2 icons instead of 4), removed inline color prop
- CheckboxInput component - Updated to pass props correctly to controlled Checkbox, cleaner prop spreading
- Checkbox recipe - Added color and _error styles to container slot (moved from inline props)
- Controlled components philosophy - New documentation explaining design system rationale
- Added event helper types to Checkbox for cleaner consumer API
- Updated stories

src/components/CheckBox/CheckBox.tsx
```diff
@@ -1,50 +1,87 @@
...
+  id?: string;
+  /** Display indeterminate state (partially checked) */
+  indeterminate?: boolean;
+  /** Disable the checkbox */
+  disabled?: boolean;
+  /** Display error state */
+  error?: boolean;
+} & Omit<BoxProps, 'checked' | 'onChange' | keyof CheckboxVariantProps> &
+  CheckboxVariantProps &
+	AriaAttributes;
```
Etwigg: Move to Box?

src/components/CheckboxInput/CheckboxInput.tsx
shaunrfox: Because it's the wrapper element, we need to merge the styles from the recipe

docs/controlled-components-philosophy.md
Etwigg, line 1: What's this docs folder for? This should be a storybook doc page, right?

src/components/CheckBox/CheckBox.tsx
```diff
@@ -1,47 +1,86 @@
 import { Box, type BoxProps } from '../Box';
-import { Label } from '../Label';
 import { checkbox, type CheckboxVariantProps } from '@styled-system/recipes';
 import { Icon } from '../Icon';
-import { AriaAttributes } from 'react';
+import { AriaAttributes, ChangeEventHandler } from 'react';
```
Etwigg: Can get rid of AriaAttributes

src/components/CheckBox/CheckBox.tsx
```diff
@@ -1,47 +1,86 @@
...
+} & Omit<BoxProps, 'checked' | 'onChange' | keyof CheckboxVariantProps> &
+  CheckboxVariantProps;
 
-export const CheckBox: React.FC<CheckBoxProps> = ({
+/**
+ * Helper type for checkbox change events
+ * @example
+ * const handleChange: CheckboxChangeHandler = (e) => setChecked(e.target.checked);
+ */
+export type CheckboxChangeEvent = React.ChangeEvent<HTMLInputElement>;
```
Etwigg: I don't think I understand why we're doing this?

docs/controlled-components-philosophy.md
```diff
@@ -0,0 +1,578 @@
...
+```typescript
+// Parent controls the state explicitly
+const [isChecked, setIsChecked] = useState(false);
+
+<Checkbox
+  name="terms"
+  checked={isChecked}
+  onChange={(e) => setIsChecked(e.target.checked)}
+/>
+```
```
Etwigg: This example seems weird to me. You're basically doing the same thing in both. You just haven't defined handleChange in the first. It would be a controlled component (or the same as the second example) if handleChange set the state, right?
Uncontrolled would be using refs?

src/components/CheckBox/CheckBox.tsx
```diff
@@ -1,47 +1,86 @@
...
+
+/**
+ * Helper type for checkbox change handler functions
+ * @example
+ * const handleChange: CheckboxChangeHandler = (e) => setChecked(e.target.checked);
+ */
+export type CheckboxChangeHandler = (e: CheckboxChangeEvent) => void;
```
Etwigg: or this

src/components/CheckBox/CheckBox.tsx
```diff
@@ -1,47 +1,86 @@
...
       <Box
         as="input"
         type="checkbox"
         className={input}
         name={name}
         id={id}
-        aria-label={name}
-        {...props}
+        checked={checked}
+        onChange={onChange}
```
Etwigg: We don't need to explicitly list checked and onChange here. You can just toss those on for free because of the Box component.

src/components/CheckBox/CheckBox.tsx
```diff
@@ -1,47 +1,86 @@
...
+ * />
+ */
+
+export const Checkbox: React.FC<CheckboxProps> = ({
   indeterminate,
   error,
   id,
   name,
+  checked,
+  onChange,
```
Etwigg: Don't need these here.

src/components/CheckBox/CheckBox.tsx
```diff
@@ -1,47 +1,86 @@
...
+  /** Unique identifier for the checkbox */
+  id?: string;
+  /** Display indeterminate state (partially checked) */
+  indeterminate?: boolean;
+  /** Disable the checkbox */
+  disabled?: boolean;
+  /** Display error state */
+  error?: boolean;
+} & Omit<BoxProps, 'checked' | 'onChange' | keyof CheckboxVariantProps> &
+  CheckboxVariantProps;
```
Etwigg: Are we planning on reformatting all of these? This is just flipping the code style backwards. Is there a reason to do it this way?

src/components/CheckBox/CheckBox.tsx
```diff
@@ -1,47 +1,86 @@
...
   ...props
 }) => {
   const { container, input, indicator } = checkbox({});
 
+  // Determine which icon to render based on state
+  const iconName = indeterminate
+    ? 'checkbox-indeterminate'
+    : checked
+      ? 'checkbox-checked'
+      : 'checkbox';
```
Etwigg: Good change!

src/components/CheckBox/CheckBox.tsx
```diff
@@ -1,47 +1,86 @@
...
-      color={error ? 'red.50' : { base: 'slate.90', _dark: 'slate.0' }}
-    >
+    <Box className={container} {...(error && { 'data-error': true })}>
       <Box
         as="input"
         type="checkbox"
         className={input}
         name={name}
         id={id}
-        aria-label={name}
```
Etwigg, line 37: Don't we still need aria-label for this? Or is having it on the parent (CheckboxInput) enough to pass accessibility? 
If we try to use the checkbox outside of the CheckboxInput are we going to run into issues? Or are we saying it'll always have a parent that will have the aria-label?

src/components/Checkbox/index.tsx
```diff
@@ -1 +1,6 @@
-export { CheckBox } from "./CheckBox";
\ No newline at end of file
+export {
+  Checkbox,
+  type CheckboxProps,
+  type CheckboxChangeEvent,
+  type CheckboxChangeHandler,
```
Etwigg, line 5: These can go

src/components/CheckboxInput/CheckboxInput.stories.tsx
```diff
@@ -0,0 +1,439 @@
...
+    const handleChange: CheckboxChangeHandler = (e) =>
+      setChecked(e.target.checked);
+
+    return (
+      <CheckboxInput
+        name="terms"
+        id="terms"
+        checked={checked}
+        onChange={handleChange}
+      >
```
Etwigg, line 96: This is the example of an uncontrolled component from your doc file.
(I think it's controlled)

src/components/CheckboxInput/CheckboxInput.stories.tsx
```diff
@@ -0,0 +1,439 @@
...
+import { Text } from '../Text';
+import { Button } from '../Button';
+
+const meta: Meta<typeof CheckboxInput> = {
+  title: 'Components/CheckboxInput',
+  component: CheckboxInput,
+  parameters: {
+    layout: 'centered',
+  },
+  // Only include props explicitly defined in CheckboxInputProps
```
Etwigg, line 17: We're pulling 'as' from somewhere too. CheckboxInput doesn't have that as a prop since it's just a container for the Checkbox itself.

src/components/CheckboxInput/CheckboxInput.stories.tsx
Etwigg, line 1: Some weird stuff on the doc page 
image

src/components/CheckboxInput/CheckboxInput.stories.tsx
```diff
@@ -0,0 +1,439 @@
...
+          >
+            Option 4
+          </CheckboxInput>
+        </Box>
+      </Box>
+    );
+  },
+};
+
+export const ExFormIntegration: Story = {
```
Etwigg, line 272: This isn't how we're going to be handling forms though, right? We've got form components.

src/components/CheckboxInput/CheckboxInput.stories.tsx
```diff
@@ -0,0 +1,439 @@
...
+    expect(checkbox).toBeChecked();
+
+    // Test spacebar to uncheck
+    await userEvent.keyboard(' ');
+    expect(checkbox).not.toBeChecked();
+  },
+  parameters: {
+    a11y: {
+      config: {
+        rules: [
```
Etwigg, line 381: Did you setup https://storybook.js.org/docs/writing-tests/accessibility-testing ?
I thought we were supposed to have an 'Accessibility' tab on the interaction panel that I'm not seeing.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -1,35 +1,43 @@
...
 import { FC, ReactNode } from 'react';
-import { Box, BoxProps } from '../Box';
+import { cx } from '@styled-system/css';
+import { BoxProps } from '../Box';
 import { Label } from '../Label';
-import { CheckBox } from '../CheckBox/CheckBox';
+import { Checkbox } from '../Checkbox';
+import {
+  checkboxInput,
+  type CheckboxInputVariantProps,
```
Etwigg, line 9: Is this ok to import if we don't have any variants?

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -1,35 +1,43 @@
...
   error,
   indeterminate,
   ...props
-}: CheckBoxInputProps) => {
+}: CheckboxInputProps) => {
+  const [className, otherProps] = splitProps(props);
   return (
-    <Label htmlFor={id}>
-      <CheckBox
+    <Label className={cx(checkboxInput(), className)} {...otherProps} htmlFor={id}>
```
Etwigg: checkboxinput recipe applies to label?

Etwigg: I'm seeing the checkbox disappear when trying to check it.
I think there's something wrong with the css here, or changing the props or both.

https://github.com/user-attachments/assets/f8760946-863d-40e6-8a49-32deccf06f15

Etwigg (changes requested):

Etwigg:
@shaunrfox 
Just want to make sure this is still on your radar.

shaunrfox:
> @shaunrfox Just want to make sure this is still on your radar.

@Etwigg Sorry, I need to revisit this. I'll take a look at it today.

docs/controlled-components-philosophy.md
shaunrfox: moved to storybook

src/components/CheckboxInput/CheckboxInput.stories.tsx
shaunrfox: I'll add this in this PR

src/components/CheckBox/CheckBox.tsx
shaunrfox: We only need an aria-label if there's not a proper <label> present. Having both is a no-no. It doesn't need to be explicit here, though, because you can pass it as a generic prop any time. This should be an exception case.

src/components/Checkbox/index.tsx
shaunrfox: This is part of Checkbox being a controlled component, having explicit states instead of deriving the state from the dom. These allow consumers to hook into and affect the state.

src/components/CheckboxInput/CheckboxInput.tsx
shaunrfox: Yes. It's a good pattern to be consistent and it doesn't harm anything. That type is automatically generated by Panda.

docs/controlled-components-philosophy.md
shaunrfox: Uncontrolled means the state is derived from the dom. Controlled means React state is the source of truth.

src/components/CheckboxInput/CheckboxInput.tsx
shaunrfox: this is fixed now

atkinsdavid (approved):
Not sure if there are more to-dos here, but looks okay to me.

PR #117, opened 2025-11-13
refactor: converts radio to controlled component. adds stories.

shaunrfox:
**Radio Component - Controlled Pattern Implementation**

- **Radio.tsx:** Converted to controlled component requiring checked, onChange,
and value props; removed Label wrapper; implemented conditional icon rendering (2 icons instead
of 3); removed problematic aria-label; added helper types (RadioChangeEvent, RadioChangeHandler)
- **Radio/index.tsx:** Exported helper types for consumers

**RadioInput Component - Controlled Pattern Implementation**

- **RadioInput.tsx:** Made controlled (requires checked, onChange, value); removed
 htmlFor; removed Text wrapper; fixed props spreading order; passes error as prop instead of data
 attribute
- **RadioInput.stories.tsx:** New comprehensive Storybook stories with Default, All
 States, 3 interactive examples, and 2 A11y test stories

**Recipe Updates**

- **radio.ts:** Added semantic error token (error.default instead of red.50); added
 container color with error state
- **radioinput.ts:** Minor recipe adjustments

**App.tsx Demo Updates**

- **App.tsx:** Updated Radio and RadioInput demos to use controlled pattern with
state management


Etwigg:
@shaunrfox You're duplicating changes from your preset branch.
This shouldn't be pointed at main.

shaunrfox:
Let's wait on this until the Text, Checkbox, and Tokens PRs are merged, then I'll take another look.

PR #119, opened 2025-12-16
feat: add new props for menu component

vidhi-parikh:
Summary: Added loading and onClose prop for the Menu component.


src/components/Menu/Menu.tsx
```diff
@@ -114,122 +149,140 @@ export const Menu: React.FC<MenuProps> = ({
...
-                        color={{ base: 'slate.90', _dark: 'slate.0' }}
-                      >
-                        {item?.iconName && (
-                          <Icon name={`${item?.iconName as IconNamesList}`} />
+        {loading ? (
+          <Box
+            display="flex"
+            alignItems="center"
+            justifyContent="center"
+            py="4"
```
smitjp150622: Please manage the CSS using slots. In our design system, we don’t use inline CSS.

src/components/Menu/Menu.tsx
```diff
@@ -34,6 +37,8 @@ export type MenuProps = Omit<BoxProps, keyof MenuVariantProps> &
   };
 
 export const Menu: React.FC<MenuProps> = ({
+  loading,
+  onClose,
```
Etwigg, line 42: I was thinking about this more and thought about using the dialog element for our menu options? We would get onClose for free this way and not have to worry about using it as a prop.
@shaunrfox any thoughts about that?

src/components/Menu/Menu.tsx
```diff
@@ -63,6 +68,36 @@ export const Menu: React.FC<MenuProps> = ({
...
+      if (event.key === 'Escape') {
+        onClose?.();
+      }
+    };
+
+    document.addEventListener('keydown', handleKeyDown);
+    return () => {
+      document.removeEventListener('keydown', handleKeyDown);
+    };
+  }, [onClose]);
```
Etwigg: If we don't go the dialog route then I do think this would need to be in a utility function like you mentioned from the beginning. I don't think I understood how verbose this would be.

Etwigg (changes requested):

Etwigg:
@vidhi-parikh 
You still need to make the utility function for onClose here, right?

vidhi-parikh:
> @vidhi-parikh You still need to make the utility function for onClose here, right?

Yess. So sorry, I got busy with progress-bar figma that it slipped out of my head. Will make these changes. Thanks for the reminder.

Etwigg (approved):

smitjp150622 (approved):

smitjp150622 (approved):
🚀 LGTM

PR #124, opened 2026-01-06
refactor: card refactor + storybook fixes

shaunrfox:
Primary thing here is the Card component + stories. (I don't think these changes break the interface, but I haven't looked at the implementation in the website code)

Most other storybook stuff is fixing an issue with the docs in-page navigation.


shaunrfox:
This PR has been gobbled up by https://github.com/Cetec-ERP/Cetec-Design-System/pull/126 and is no longer needed.

PR #125, opened 2026-01-07
build!: changed build to make it easier to consume the library

shaunrfox:


Etwigg:
I'm struggling to get this to work.
Getting build fails because of container, it's still not seeing any of the exports, etc.
We can look at it tomorrow.

Etwigg (approved):

PR #126, opened 2026-01-12
refactor!: button refactor

shaunrfox:
Button + IconButton updates:
- Now use slot recipes + semantic colors
- Now use forwardRef
- Now receive icons as props
- Removed 'ButtonContent' component
- Added 'xl' size
- Size variants now use short size names (sm, md, lg, xl)

Spinner updated to use explicit colors instead of automatic inverse filter

Removed global _disabled opacity declaration (now handled by semantic colors)


src/components/Box/Box.tsx
```diff
@@ -24,3 +29,15 @@ export const Box: React.FC<BoxProps> = ({ as = 'div', ...props }) => {
     ...otherProps,
   });
 };
+
+// React 19+: ComponentPropsWithRef<ElementType> is recommended as refs are now passed as props in function components.
```
shaunrfox, line 33: @Etwigg As I was looking into `ref` and `onClick`, this came up as the new way in React 19.

src/components/Button/Button.tsx
```diff
@@ -1,44 +1,79 @@
...
-  variant,
-  size,
-  href,
-  children,
-  loading,
-  disabled,
-  type = 'button',
-  ...props
-}: ButtonProps) => {
+export const Button = (props: ButtonProps) => {
```
shaunrfox, line 21: From my robot, Claude:

Yes, it's still a functional component. FC<> is not required.

Both are valid functional components:

// Current - recommended modern approach
export const Button = (props: ButtonProps) => { ... }

// Also valid, but unnecessary
export const Button: FC<ButtonProps> = (props) => { ... }

FC (FunctionComponent) is considered legacy/unnecessary in modern React TypeScript because:

1. It used to implicitly add children to props, which caused confusion (removed in React 18)
2. It doesn't add type safety - typing props directly on the parameter achieves the same thing
3. It makes generics awkward if you ever need them
4. The return type is inferred correctly by TypeScript anyway

The React TypeScript community has largely moved away from FC. Your current pattern is the recommended approach - it's cleaner and more explicit about what the component accepts.

src/components/Button/Button.tsx
```diff
@@ -1,44 +1,79 @@
...
       {...(href ? { href } : { type })}
       {...otherProps}
-      aria-label={children}
+      {...(trulyDisabled &&
+        href && {
+        onClick: (e: React.MouseEvent<HTMLAnchorElement>) =>
+          e.preventDefault(),
+      })}
     >
-      <ButtonContent loading={!!loading}>{children}</ButtonContent>
```
shaunrfox, line 41: ditched the extra component

src/components/Card/Card.tsx
```diff
@@ -10,26 +10,49 @@ export type CardProps = Omit<BoxProps, keyof CardVariantProps> &
...
+    interactive,
+    ...rest
+  } = props;
+  const [className, otherProps] = splitProps(rest);
+
+  // Determine if card should be interactive based on props
+  const isInteractive = interactive || Boolean(href) || Boolean(onClick);
+
+  // Determine the element to render
+  let asComponent: React.ElementType = as || 'div';
```
shaunrfox: left `as` explicitly in the props because of this logic

src/components/Icon/Icon.tsx
```diff
@@ -26,7 +26,7 @@ export type IconProps = Omit<BoxProps, 'size'> &
 export const Icon: React.FC<IconProps> = ({
   name,
   size = '24',
-  fill = 'current',
+  fill,
```
shaunrfox, line 27: This might have downstream impact, but it needs doing. It was preventing us from being able to style the icon reliably.

src/components/Spinner/Spinner.tsx
```diff
@@ -4,18 +4,21 @@ import { cx } from '@styled-system/css';
 import { splitProps } from '~/utils/splitProps';
 
 export type SpinnerProps = Omit<BoxProps, keyof SpinnerVariantProps> &
-  SpinnerVariantProps;
+  SpinnerVariantProps & {
+    inverse?: boolean;
```
shaunrfox, line 8: The css filter that was previously in the recipe proved unreliable in some situations, so this is now explicit.

src/contexts/ThemeContext.tsx
```diff
@@ -7,11 +7,11 @@ interface ThemeContextType {
   setTheme: (theme: Theme) => void;
 }
 
-//const THEME_STORAGE_KEY = 'cetec-theme-preference';
+const THEME_STORAGE_KEY = 'cetec-theme-preference';
```
shaunrfox, line 10: Added this back so that it's explicit and sticks between refreshes.

src/recipes/button.ts
```diff
@@ -1,220 +1,390 @@
...
+  slots: ['container', 'icon'],
+  base: buttonBaseStyles,
   variants: {
     ...buttonVariants,
     size: {
-      medium: {
-        fontSize: '16',
-        py: '3',
-        px: '10',
+      md: {
```
shaunrfox, line 188: We've been inconsistent with our tshirt-size naming conventions. Migrating all instances to the shorthand syntax.

src/recipes/index.ts
```diff
@@ -13,9 +12,10 @@ export { cardRecipe } from './card';
 export { breadcrumbsRecipe } from './breadcrumbs';
 export { tagRecipe } from './tag';
 export { radioInputRecipe } from './radioinput';
-export { checkboxInputRecipe } from './checkboxInput';
+export { checkboxInputRecipe } from './checkboxinput';
```
shaunrfox: This was from a build error. I think we're running into Mac + Git issues with case-only changes... Not sure how to remedy that.

src/storybook/docs/componentStandards.mdx
```diff
@@ -0,0 +1,113 @@
...
+import { Flex } from '@styled-system/jsx';
+import { Divider } from '~/components/Divider';
+
+<Meta title="Guides/ Component Standards" />
+
+<Unstyled>
+
+<Flex direction="column" gap="12">
+
+# Component standards
```
shaunrfox, line 11: Just a starting place to capture decisions we've made or research we've done.

src/styles/utilities/globalStyle.ts
```diff
@@ -13,7 +13,6 @@ export const globalCss = defineGlobalStyles({
       outlineColor: { base: 'slate.90', _dark: 'slate.0' },
     },
     _disabled: {
```
shaunrfox, line 15: This is largely being replaced with our new semantic tokens. Removed from globalStyle, but it can still be used at the component level if needed.

src/recipes/button.ts
atkinsdavid: > tshirt-size naming conventions

😆

atkinsdavid (dismissed):

src/components/Button/Button.tsx
```diff
@@ -1,44 +1,79 @@
...
+export const Button = (props: ButtonProps) => {
+  const {
+    variant,
+    size,
+    href,
+    iconBefore,
+    iconAfter,
+    children,
+    loading,
+    disabled,
```
Etwigg, line 30: we don't need this as an explicit prop, right? It's one we get for free.

atkinsdavid:
Button variant for "destructive"

src/components/Button/Button.tsx
```diff
@@ -1,44 +1,79 @@
...
+    type = 'button',
+    ...rest
+  } = props;
+  const classes = button({
+    variant,
+    size,
+    iconBefore: Boolean(iconBefore),
+    iconAfter: Boolean(iconAfter),
+  });
```
Etwigg, line 39: I don't think I like splitting it up like this? Maybe I just don't understand what we're doing below with classes.container.

src/components/Button/Button.tsx
```diff
@@ -1,44 +1,79 @@
...
 
-  const [className, otherProps] = splitProps(props);
-
   return (
     <Box
       as={href ? 'a' : 'button'}
       disabled={trulyDisabled}
       aria-disabled={trulyDisabled}
-      className={cx(button({ variant, size }), className)}
+      aria-label={children}
```
Etwigg: This seems like it won't work, right? Children could be anything.

src/components/Button/Button.tsx
```diff
@@ -1,44 +1,79 @@
...
+      className={cx(classes.container, className)}
       {...(href ? { href } : { type })}
       {...otherProps}
-      aria-label={children}
+      {...(trulyDisabled &&
+        href && {
+        onClick: (e: React.MouseEvent<HTMLAnchorElement>) =>
+          e.preventDefault(),
+      })}
```
Etwigg: I don't think I like the trulyDisabled thing (I know that's not something you added, but if we're refactoring it, maybe it's worth discussing).
I feel like we should be using the normal 'disabled' that comes with the html element, and then have an 'isLoading' or something to handle the loading state.
What do y'all think? @shaunrfox @atkinsdavid

src/components/Button/Button.tsx
```diff
@@ -1,44 +1,79 @@
...
+            top="0"
+            left="0"
+            right="0"
+            bottom="0"
+            placeItems="center"
+          >
+            <Spinner size="sm" />
+          </Grid>
+        )}
+      </>
```
Etwigg: Getting rid of the ButtonContent means all of this has to be done with inline styles now?
I feel like we should either be using a slot recipe or keep the ButtonContent component.
Do you think inline styles are better than that in this case?

src/components/Card/Card.tsx
```diff
@@ -1,35 +1,58 @@
...
 
-export type CardProps = Omit<BoxProps, keyof CardVariantProps> &
+export type CardProps = BoxProps &
   CardVariantProps & {
     href?: string;
     children?: string | ReactNode;
     grabbed?: boolean;
     disabled?: boolean;
+    interactive?: boolean;
   };
```
Etwigg: Not keeping Omit for card?

src/components/Card/Card.tsx
```diff
@@ -1,35 +1,58 @@
...
+  const {
+    as,
+    variant,
+    href,
+    onClick,
+    children,
+    disabled,
+    grabbed,
+    interactive,
+    ...rest
```
Etwigg: I think we've got some inconsistency here?
For the button, we're not using 'as'. We use 'type=' then we implicitly set as based on href being passed in.
We also don't need onClick here, right?
Same with disabled, it's just handled by the html.
Should grabbed be a conditionial instead of a prop and handled in the recipe?

src/components/Card/Card.tsx
Etwigg: Why do we need this?
We have a pattern in button that achieves the same thing, right?

src/components/Card/Card.tsx
```diff
@@ -1,35 +1,58 @@
...
 
   return (
     <Box
-      as={href ? 'a' : 'button'}
-      disabled={disabled}
-      aria-disabled={disabled}
+      as={asComponent}
       data-grabbed={grabbed}
-      className={cx(card({ variant }), className)}
+      className={cx(card({ variant, interactive: isInteractive }), className)}
```
Etwigg, line 40: We're not following the pattern in button here either. Do we want to define classes before we do the cx or do we want to do it inline? I think we need to be consistent.

src/components/Heading/Heading.tsx
```diff
@@ -9,19 +9,20 @@ export type HeadingProps = Omit<TextProps, keyof HeadingVariantProps> & HeadingV
 };
 
 export const Heading: React.FC<HeadingProps> = (
```
Etwigg: Thought we were getting rid of FC

src/components/IconButton/IconButton.tsx
```diff
@@ -18,42 +18,62 @@ import { splitProps } from '~/utils/splitProps';
  */
 export type IconButtonProps = BoxProps &
   IconButtonVariantProps & {
```
Etwigg: Did we lose Omit here at some point? Was that on purpose?

src/components/IconButton/IconButton.tsx
```diff
@@ -18,42 +18,62 @@ import { splitProps } from '~/utils/splitProps';
...
+    iconName: IconNamesList;
     href?: string;
     loading?: boolean;
-    children: ReactElement<typeof Icon>;
```
Etwigg, line 23: Why'd we remove children?

src/components/IconButton/IconButton.tsx
```diff
@@ -18,42 +18,62 @@ import { splitProps } from '~/utils/splitProps';
...
+    size,
+    href,
+    loading,
+    disabled,
+    type = 'button',
+    ...rest
+  } = props;
+  const classes = iconButton({ variant, size });
+  const [className, otherProps] = splitProps(rest);
+  const trulyDisabled = loading || disabled;
```
Etwigg, line 41: See other comments on trulyDisabled

src/components/IconButton/IconButton.tsx
```diff
@@ -18,42 +18,62 @@ import { splitProps } from '~/utils/splitProps';
...
   return (
     <Box
       as={href ? 'a' : 'button'}
-      disabled={isDisabled}
-      aria-disabled={isDisabled}
+      disabled={trulyDisabled}
+      aria-disabled={trulyDisabled}
       aria-label={`icon-button`}
-      className={cx(iconButton({ variant, size }), className)}
+      className={cx(classes.container, className)}
```
Etwigg: Do we have '.container' for iconButton? I don't see changes to the recipe.

src/components/IconButton/IconButton.tsx
```diff
@@ -18,42 +18,62 @@ import { splitProps } from '~/utils/splitProps';
...
+            top="0"
+            left="0"
+            right="0"
+            bottom="0"
+            placeItems="center"
+          >
+            <Spinner size="sm" />
+          </Grid>
+        )}
+      </>
```
Etwigg, line 76: This is going to super-break any iconButtons we're using anywhere, since we're not using children anymore.

src/components/ThemeSwitcher/ThemeSwitcher.tsx
```diff
@@ -1,8 +1,7 @@
 import { IconButton } from '~/components/IconButton';
-import { Icon } from '~/components/Icon';
 import { useTheme } from '~/contexts/ThemeContext';
 
-export const ThemeSwitcher: React.FC = ({}) => {
+export const ThemeSwitcher: React.FC = ({ }) => {
```
Etwigg: FC

src/components/Button/Button.tsx
```diff
@@ -1,44 +1,79 @@
...
+          <Grid
+            position="absolute"
+            top="0"
+            left="0"
+            right="0"
+            bottom="0"
+            placeItems="center"
+          >
+            <Spinner size="sm" />
+          </Grid>
```
Etwigg: Would it be bad to include the whole grid component in the spinner? Small concerns about having to duplicate this everywhere we use spinner.

src/recipes/spinner.ts
```diff
@@ -6,37 +6,47 @@ const spinnerBase = {
   borderWidth: '3',
   borderStyle: 'solid',
   borderColor: 'transparent',
-  borderTopColor: 'slate.90',
-  borderBottomColor: 'slate.90',
+  borderTopColor: 'icon/60',
+  borderBottomColor: 'icon/60',
```
Etwigg: What's with the '/'? I'd like to stay consistent with token naming schemes if possible.

src/storybook/tokens/components/SizeToken.tsx
```diff
@@ -38,7 +38,7 @@ export const wrapperStyles = css({
   flexDirection: 'column',
   bg: 'transparent',
   gap: '4',
-  rounded: '4',
+  rounded: '2',
```
Etwigg, line 41: 🤨

Etwigg (commented):
Just some questions.. I think mostly smaller/nit-pick stuff. This is going to require lots of patching for the people currently working on things...
I think I might be wrong about the 'disabled' prop... I remember talking about that at some point and we might've said it was needed because of aria-disabled or something....

src/components/Card/Card.tsx
shaunrfox: Yeah, you're right. Removed.

src/components/Button/Button.tsx
shaunrfox: This isn't adding any props, it's just destructuring them so we can access them directly in the component below.

src/components/Button/Button.tsx
shaunrfox: I think this is a pattern for slot recipes: https://panda-css.com/docs/docs/concepts/slot-recipes#using-the-recipe-1

src/components/Button/Button.tsx
shaunrfox: Yeah, I'm not sure when that was added. I'll remove.

src/components/Button/Button.tsx
shaunrfox: That's not a bad idea. We'd probably want to set it up as a variant of spinner, since we'll want it to be able to be inline as well. I'd like to think about it some more before we try this, though.

src/components/Card/Card.tsx
shaunrfox: I think this is the standard way to do it for a basic recipe (non-slot), which we're doing in lots of other components.

src/components/Heading/Heading.tsx
shaunrfox: Yeah, we are. I haven't done it for all the components yet, but I'll update this one.

src/components/IconButton/IconButton.tsx
shaunrfox: I removed the children to better align with the Button component's `iconBefore` and `iconAfter`. This helps make all the padding/positioning handled in the same way for any button components and is needed to handle icon sizing through the recipe. 

This is also a convention I plan to use for a lot of upcoming components, ex: inputs, chips, tags.

src/components/IconButton/IconButton.tsx
shaunrfox: Yes, it will be a one-time pain 😬

src/recipes/spinner.ts
shaunrfox: This is not a new convention for tokens, but a Panda feature we haven't used much yet. It's a shorthand opacity modifier: https://panda-css.com/docs/docs/concepts/color-opacity-modifier

src/storybook/tokens/components/SizeToken.tsx
shaunrfox: rude

src/components/IconButton/IconButton.tsx
shaunrfox: I didn't see another comment about this...

src/components/Button/Button.tsx
shaunrfox: Yeah, I agree. That's felt a little weird to me. I'll try something else and see what y'all think.

src/components/Card/Card.tsx
shaunrfox: I think I brought this back.

src/components/Button/Button.tsx
Etwigg: Sorry, I was referring to the disabled prop specifically here.

src/components/Button/Button.tsx
Etwigg: ok, cool. I'll add it to the list of refactors.

src/components/Card/Card.tsx
Etwigg: Ok, so slot recipe should be broken out and normal should be inline?
I'll make note of that. We can add it to our docs.

src/components/IconButton/IconButton.tsx
Etwigg: You responded to it

src/components/IconButton/IconButton.tsx
Etwigg: Let's just make sure to communicate in the PR description so we can pick it up with the changelog.

src/recipes/spinner.ts
Etwigg: oh, neat.

src/components/Button/Button.tsx
```diff
@@ -38,22 +38,25 @@ export const Button = (props: ButtonProps) => {
...
   });
   const [className, otherProps] = splitProps(rest);
-  const trulyDisabled = loading || disabled;
 
   return (
     <Box
       as={href ? 'a' : 'button'}
-      disabled={trulyDisabled}
-      aria-disabled={trulyDisabled}
-      aria-label={children}
```
Etwigg, line 48: We still need aria-label, right?

src/components/Button/Button.tsx
```diff
@@ -38,22 +38,25 @@ export const Button = (props: ButtonProps) => {
...
-        })}
+      {...(loading && {
+        'aria-busy': true,
+        'aria-live': 'polite',
+      })}
+      {...(disabled && {
+        disabled: true,
+        'aria-disabled': true,
+        ...(href && {
+          tabIndex: -1,
```
Etwigg, line 56: We don't need tabIndex here. If it's got the disabled attribute then it's removed from keyboard nav.
This works because it's actually a button element.

Etwigg (approved):

src/components/IconButton/IconButton.tsx
```diff
@@ -16,44 +16,64 @@ import { splitProps } from '~/utils/splitProps';
...
+  const [className, otherProps] = splitProps(rest);
+  const trulyDisabled = loading || disabled;
 
   return (
     <Box
       as={href ? 'a' : 'button'}
-      disabled={isDisabled}
-      aria-disabled={isDisabled}
+      disabled={trulyDisabled}
+      aria-disabled={trulyDisabled}
```
atkinsdavid, line 47: madly...deeply

atkinsdavid (approved):

PR #127, opened 2026-01-14
refactor!: panda build output optimization

shaunrfox:
These changes were initially pursued in order to allow non-react pages in the app repo to access to our design system styles (ex: template toolkit). As we explored options, we realized that the `staticCss` was bloating our library unnecessarily.

- Updated the build to only include the `staticCss` when we want it (Storybook, not prod)
- Updated build to generate CSS + some other helpers that can be used by consuming apps as needed


panda.config.ts
```diff
@@ -1,17 +1,52 @@
...
+            fontSize: ['*'],
+            fontWeight: ['*'],
+            lineHeight: ['*'],
+            letterSpacing: ['*'],
+          },
+          conditions: ['light', 'dark'],
+        },
+      ],
+    }
+  : { css: [] };
```
atkinsdavid:
```suggestion
  : {};
```

Does the empty `css:` block there cause any strange behavior? Or would it be better to omit the block entirely if we don't need it?

atkinsdavid (dismissed):

shaunrfox:
| Old dist | New dist |
|--------|--------|
| dist-old | revised-dist |

panda.config.ts
shaunrfox: Yeah, we can probably just omit the block. I'll look into that.

src/components/Icon/Icon.tsx
```diff
@@ -4,10 +4,9 @@ import { cx } from '@styled-system/css';
 import { IconNamesList } from './icons';
 import { icon } from '@styled-system/patterns';
 import { numericSizes } from '~/styles/primitives';
-import { splitProps } from '~/utils/splitProps';
-import { ConditionalValue } from '@styled-system/types';
-import { WithEscapeHatch } from '@styled-system/types/prop-type';
 import { ColorToken } from '@styled-system/tokens';
+import { splitProps } from '~/utils/splitProps';
+// import { ColorToken } from '@styled-system/tokens';
```
atkinsdavid:
```suggestion
```

atkinsdavid (dismissed):

shaunrfox:
Design System changes that need to be updated in consuming apps

## size prop values – name changes

Examples:
- small → sm
- large → lg
- default → md

## Icons

- Removed `color: 'current'` in recipe
- Replace every instance of <Icon> with correct fill=...

## All imports

- `import "..." from "@cetec-design-system/...";`
- `import "..." from "cetec-design-system";`

atkinsdavid:
> * Replace every instance of  with correct fill=...

Every instance of...what?

Etwigg:
> > * Replace every instance of  with correct fill=...
> 
> Every instance of...what?

Every Icon, right?

Etwigg:
> Old dist	New dist
> dist-old	revised-dist

Should the mcp stuff be in the package?

src/types/panda-augment.ts
Etwigg, line 1: What was the reason for this file?

src/types/panda-shim.ts
Etwigg, line 1: Does this not overlap with panda-augment.ts?

src/cetec-preset.ts
```diff
@@ -56,7 +52,7 @@ const theme = {
   },
 };
 
-export const cetecPreset = definePreset({
+export const cetecPreset: any = definePreset({
```
Etwigg: There's not a panda type for this?

src/index.ts
```diff
@@ -24,3 +24,22 @@ export { Breadcrumbs } from './components/Breadcrumbs';
...
+  SlotRecipeVariantRecord,
+  SlotRecipeConfig,
+  SystemStyleObject,
+  GlobalStyleObject,
+  CompositionStyles,
+  PatternProperties,
+  PatternConfig,
+  Parts,
+  Panda,
+} from './types/panda-augment';
```
Etwigg, line 45: panda-shims.ts is a different thing we don't want to export I guess?
Do we have to import this here before exporting?

src/index.ts
```diff
@@ -1,4 +1,4 @@
-export { Box } from './components/Box';
+export { Box, type BoxProps } from './components/Box';
```
Etwigg, line 1: Why only export BoxProps here?

Etwigg (dismissed):
This doesn't seem too crazy.

Etwigg:
@shaunrfox @atkinsdavid 
Merge if confident. I just left some questions/comments.

src/cetec-preset.ts
shaunrfox: Yeah, good call.

src/index.ts
shaunrfox: Because we use BoxProps as the main polymorphic connector to the design system, exporting this will allow us to use the props in other external components we might need to make.

src/types/panda-augment.ts
shaunrfox: Problem it solves: When consumers install the design system, they shouldn't need @pandacss/types as a dependency. But the generated type declarations (.d.ts files) from your build would normally reference Panda CSS types like RecipeConfig, SystemStyleObject, etc. — causing TypeScript errors in consumer projects that don't have @pandacss/types installed.

What it does: It provides minimal, structurally-compatible type shims (Record<string, unknown> based) that stand in for the real Panda CSS types. These are:
- Exported from src/index.ts (lines 33–45) as part of the public API
- Documented in src/storybook/docs/build-reasoning.mdx as being added to "extend Panda type generation without leaking @pandacss/types into consumer projects"

It was introduced in commit 2171769: "Finally working with no ts errors in mini test app" — this was the fix that resolved TypeScript errors when consuming the package in external packages, testing with the website repo and a mini test app.

src/types/panda-shim.ts
shaunrfox: Yeah, you're right. This is redundant now. I'll remove it.

Etwigg (approved):

atkinsdavid (approved):

PR #136, opened 2026-02-20
refactor!: global styles scoping

shaunrfox:
This abstracts out the global styles that were causing styles to leak out in consuming apps. In particular, the styles using naked tag selectors in the `globalStyles.ts` file. This should have no visible effect on the components themselves.

The tag-based styles that were removed are still available in the published package:
`/dist/reset.css`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,4 @@ storybook-static
 .project-structure.md
 .session-summary.md
 .pair-programming-session.md
+/.tmp
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -12,6 +12,7 @@ const preview: Preview = {
         dark: 'dark',
       },
       defaultTheme: 'light',
+      parentSelector: 'body',
     }),
   ],
   initialGlobals: {},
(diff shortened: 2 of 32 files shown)


atkinsdavid (dismissed):

Etwigg:
@shaunrfox 
You tested this on the website?

shaunrfox:
Tested with website and all looks good. Will need to update this PR once this is released: https://github.com/Cetec-ERP/Cetec-Website/pull/144

atkinsdavid:
Tested in the App, seems to solve the issue with the style leakage. 💧

Etwigg:
This fixes the panda leaking into the old version, but not the old version leaking into the panda, right?

atkinsdavid:
> This fixes the panda leaking into the old version, but not the old version leaking into the panda, right?

Correct, old styles that aren't properly scoped are still going to hit new stuff, if the selectors are element based for example.

There's some work started on addressing those, but nothing ready for primetime yet.

atkinsdavid (dismissed):

atkinsdavid:
Confirmed that the most recent version of these changes are what's needed to scope styles to only new design system components in the app.

atkinsdavid (approved):

Etwigg:
> Confirmed that the most recent version of these changes are what's needed to scope styles to only new design system components in the app.

That was it?

Etwigg (approved):
the yoloist of yolos

PR #141, opened 2026-03-20
ci: add new rules for auto prerelease and release flow so we can have a stable release and a prerelease from the same branch.

Etwigg:


Etwigg:
@shaunrfox 
Here's your release stuff.
You tag everything with the relevant label. Major/Minor/Patch will make pre-releases, those + release will make a normal release (stable).

PR #147, opened 2026-03-31
feat(minor): strict data-attribute conditions, icon size variants, and conditions docs

shaunrfox:
## Summary

- **Conditions** — Replace all `[data-x]` presence checks in `conditions.ts` with `[data-x=true]` strict equality so conditional styles only fire when attributes are explicitly set to boolean `true`. Fixes a bug where `data-x={false}` would incorrectly trigger styles.
- **Icon recipe** — Refactor `size` from a direct `width` prop into a recipe variant backed by `staticCss: [{ size: ['*'] }]`, ensuring all `AllowedIconSizes` are included in the built bundle.
- **TextInput** — Add `valid`, `invalid`, and `readOnly` props with corresponding `_valid`/`_invalid` recipe styles.
- **Conditions docs** — New `Docs / Using Conditions` Storybook page covering all 80+ conditions with rendered examples grouped by category, using real design system components (`Card`, `FormField`, `TextInput`, `Chip`, `CheckboxInput`, `Button`).

## Test plan

- [ ] Verify icon `size` prop renders correctly at various sizes (e.g. `<Icon name="check" size="32" />`)
- [ ] Confirm icon size CSS classes are present in the built bundle
- [ ] Test that components using data-attribute conditions (`Chip`, `Card`, `Menu`, `FormField`) still apply hover/disabled/checked/error styles correctly
- [ ] Confirm `data-x={false}` no longer incorrectly triggers conditional styles
- [ ] Check `TextInput` with `valid`, `invalid`, and `readOnly` props renders the correct border colors
- [ ] Review the new Conditions docs page in Storybook — verify rendered demos and reference table


src/components/TextInput/TextInput.tsx
```diff
@@ -14,6 +14,9 @@ export type TextInputProps = Omit<BoxProps, keyof TextInputVariantProps> &
     iconAfter?: IconNamesList;
     error?: boolean;
     disabled?: boolean;
+    readOnly?: boolean;
```
Etwigg: Why do we need readOnly? Isn't it effectively the same as disabled?

src/recipes/recipes-regular.ts
Etwigg, line 1: something is unhappy in here!

src/recipes/recipes-slot.ts
Etwigg, line 1: Where'd this come from? Thought we moved these already.

.storybook/story-docs-style.css
```diff
@@ -134,6 +134,8 @@ code {
   border-radius: 0.25rem;
   margin-block: -0.25rem;
   mix-blend-mode: multiply;
+  white-space: wrap !important;
```
atkinsdavid, line 137: very...

src/components/TextInput/TextInput.tsx
atkinsdavid: I was wondering the same thing.

.storybook/story-docs-style.css
Etwigg: uh-oh!

src/components/TextInput/TextInput.tsx
shaunrfox: We may not have a direct need right now, but I was trying to account for conditions that did not yet have working styles.

src/recipes/recipes-slot.ts
shaunrfox: Lines got crossed. Fixed now

.storybook/story-docs-style.css
shaunrfox: Storybook docs CSS hack. Come at me, bro.

src/recipes/recipes-regular.ts
shaunrfox: fixed

src/components/TextInput/TextInput.tsx
Etwigg: if it's a condition do you need to define it here as a prop still? You can have _readOnly the condition that just sets disabled, right?

shaunrfox: Ok, yeah. Good call. "readonly" is a normal html attribute for inputs, so I updated the condition to respond to that and added the style condition to the textInput recipe.

Etwigg (approved):
full send

Etwigg:
:rocket: PR was released in [`1.2.0`](https://github.com/Cetec-ERP/Cetec-Design-System/releases/tag/1.2.0) :rocket:

PR #151, opened 2026-04-06
fix: make sure CSS for selected list items is generated

atkinsdavid:
Came up while debugging the selected item path in the app

### DIFF
--- a/src/recipes/listItem.ts
+++ b/src/recipes/listItem.ts
@@ -13,6 +13,11 @@ export const listItemRecipe = defineSlotRecipe({
     'itemDescription',
     'divider',
   ],
+  staticCss: [
+    {
+      selected: ['*'],
+    },
+  ],
   base: {
     wrapper: {
       ...globalBaseStyles,


Etwigg (commented):
We shouldn't need to do this with Shaun's fixes last week, right? What's up with this one?

shaunrfox:
> We shouldn't need to do this with Shaun's fixes last week, right? What's up with this one?

This is for a different component. David and I talked about it. These are probably going to come up occasionally as we exercise the system.

shaunrfox (approved):

atkinsdavid:
@Etwigg what would be best here for the CC stuff? Do I need to add `(patch)` to the title or something?

Etwigg:
> @Etwigg what would be best here for the CC stuff? Do I need to add `(patch)` to the title or something?

Yeah, that'll auto set the tag.

Etwigg (approved):

Etwigg:
> > We shouldn't need to do this with Shaun's fixes last week, right? What's up with this one?
> 
> This is for a different component. David and I talked about it. These are probably going to come up occasionally as we exercise the system.

I meant your changes around the buildinfo

Etwigg:
:rocket: PR was released in [`1.2.1`](https://github.com/Cetec-ERP/Cetec-Design-System/releases/tag/1.2.1) :rocket:

PR #160, opened 2026-04-28
fix: shared field and slot context layers + arbitrary slot handling

shaunrfox:
Made FormField prop propagation more reliable by adding a shared context layer.


atkinsdavid (dismissed):

Etwigg (commented):
## Findings

1. `src/recipes/textInput.ts` - **High** - Hard-coded sharp radii in `buttonSlot` creates visual defects for `before` slots

   Issue: The `buttonSlot` base style hard-codes `borderTopLeftRadius: '0'` and `borderBottomLeftRadius: '0'`.
   Impact: When a `Button` or `IconButton` is placed in the `before` (left) slot, the outer left edge of the input component will be sharp while the inner edge (touching the input) remains rounded. This is the opposite of the intended visual "flush" connection.
   Fix: Move the radii overrides to the `before` and `after` variants in the recipe so they can target the correct side.

2. `src/components/FormField/FormField.tsx` - **Medium** - Redundant prop injection alongside `FieldContext`

   Issue: `FormField` still uses `Children.map` and `cloneElement` to inject `size`, `error`, etc., into direct children, but now also wraps them in `FieldContext.Provider`.
   Impact: Components that consume `FieldContext` will receive the same values twice (via context and props), which can lead to confusion and conflicting state if props are manually overridden on the child.
   Fix: Plan a migration to remove `cloneElement` once all child components (`TextInput`, `Checkbox`, etc.) have been updated to use `useFieldContext`.

3. `src/recipes/button.ts` - **Low** - Unused slots in `buttonRecipe`

   Issue: The `buttonRecipe` defines `before`, `after`, `iconBefore`, and `iconAfter` as slots, but these keys are never used as class names in `Button.tsx`.
   Impact: Unnecessary bloat in the generated Panda CSS. In Panda, `before` and `after` are being used as *variants* that target the `mainContent` slot, so they do not need to be declared in the `slots` array unless they are also intended to be distinct DOM elements with their own classes.
   Fix: Remove unused keys from the `slots` array in `buttonRecipe`.

## Speculative Risks

- `src/components/TextInput/TextInput.tsx` - The `buttonSlot` uses `m: '-1'` to pull the button flush. If the button has a significant box-shadow or ring, this might cause clipping since the container no longer has `overflow: hidden`. Confirm that buttons with focus rings still render correctly when slotted.
- `src/components/Button/Button.tsx` - The size resolution logic `sizeProp ?? slotContext?.size ?? fieldContext?.size` is robust, but ensure that `slotContext?.size` is correctly typed. The cast `as ButtonVariantProps['size']` is safe for now but might drift if `SlotContext` is used by non-button-like owners.

shaunrfox:
@Etwigg good feedback. All fixed.

Etwigg:

PR #165, opened 2026-05-05
refactor: pass href prop to render listitem as anchor

atkinsdavid:
https://internal.cetecerpbeta.com/otd/customercomplaint/38334/edit
https://internal.cetecerpbeta.com/otd/customercomplaint/38364/edit

Some reports of users and internal people wanting the ability to right click global search results in order to open them in new tabs/windows.

Shouldn't require refactoring anywhere the component is being used, because the prop is optional.

### DIFF
--- a/src/components/List/List.stories.tsx
+++ b/src/components/List/List.stories.tsx
@@ -245,6 +245,26 @@ export const Highlighting: Story = {
   parameters: { controls: { disable: true } },
 };
 
+export const WithHrefs: Story = {
+  args: {},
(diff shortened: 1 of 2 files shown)


atkinsdavid:
Another complaint about this today (mostly noting here so I have a list of cases to follow up on when the fix gets merged in the app): https://internal.cetecerpbeta.com/otd/customercomplaint/38820/edit

Etwigg (approved):

Etwigg:

PR #170, opened 2026-06-19
feat: shadow refactor to add dropShadow property and tokens

shaunrfox:
both boxShadow and dropShadow tokens are created from the same source


src/styles/utilities/shadowDefinitions.ts
```diff
@@ -0,0 +1,72 @@
+type ShadowLayer = {
+  x: string;
+  y: string;
+  blur: string;
+  color: string;
+  spread?: string;
+};
+
+const sizeToken = (value: string) => `{sizes.${value}}`;
+const colorToken = (value: string) => `{${value}}`;
```
Etwigg: Shouldn't these be the actual token type and not just string?

Etwigg (commented):
These aren't breaking changes?

src/styles/utilities/shadowDefinitions.ts
shaunrfox: Good catch. I'll update this

shaunrfox:
> These aren't breaking changes?

I suppose they might be. But they're style-only, not functionality...

atkinsdavid (approved):

Etwigg (approved):

Etwigg:

PR #177, opened 2026-07-29
fix: ai review fixes

Etwigg:
This branch fixes four Autocomplete review findings:
Type-safe selection modes: array values now require multiple={true}. Existing AutocompleteProps<true> and <false> usage remains compatible.
Single Escape handling: removed the duplicate local Escape handler so Floating UI closes the popup and calls onOpenChange only once.
Focus restoration: clicking a token’s remove button now returns focus to the Autocomplete input.
Valid option semantics: multi-select options use decorative checkbox icons instead of nesting native checkboxes inside option buttons.
It also adds Storybook interaction coverage for Escape callback counts, token-removal focus, and checkbox-free option semantics. ListItem gained a generic decorative before slot to support the new option indicator.

### DIFF
--- a/src/components/Autocomplete/Autocomplete.stories.tsx
+++ b/src/components/Autocomplete/Autocomplete.stories.tsx
@@ -73,14 +73,17 @@ type Story = StoryObj<typeof meta>;
 
 export const Default: Story = {
   render: function DefaultRender(args) {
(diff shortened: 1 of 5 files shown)


shaunrfox (approved):
Makes sense to me

atkinsdavid (approved):

Etwigg:

PR #182, opened 2026-08-05
perf(config): optimize Playroom CSS generation

shaunrfox:
## Summary

- split Playroom static CSS generation into themed and responsive property groups
- avoid generating breakpoint variants for color properties and theme variants for layout and typography properties
- regenerate the committed Playroom stylesheet with the optimized configuration
- document the rationale, measured impact, and fallback for genuinely responsive color usage

## Why

The previous configuration applied both light/dark conditions and responsive generation to every static CSS property. That produced a large number of combinations that cannot occur in normal usage and made Playroom CSS generation unnecessarily slow.

## Impact

- measured CSS generation time drops from about 124 seconds to about 21–23 seconds
- generated stylesheet drops from 8,422,581 bytes / 259,934 lines to 4,623,430 bytes / 148,563 lines
- Playroom builds and cache misses complete substantially faster

## Validation

- `npm run playroom:build`
- regenerated output matched the committed stylesheet; working tree remained clean


Etwigg (approved):

Etwigg:

PR #183, opened 2026-08-12
feat(minor): emit data-ds-component on component roots

shaunrfox:
Adds a `data-ds-component` attribute to component roots so the ERP front end can identify design-system components in the DOM without relying on class names or test ids.

This is the base of the front-end instrumentation work — the attribute is the anchor that behavior logging and automated tests both key off of.

## What's here

- **New:** `src/utils/dsComponent.ts` — helper that resolves the attribute value from a component's display name.
- Applies `data-ds-component` on the root element of the component set (48 files, +261/−6).
- Story coverage added for `TextInput`.

## Notes

- Additive only — no API changes, no breaking changes.
- Stacked below https://github.com/Cetec-ERP/Cetec-Design-System/pull/184, which builds `data-ds-part` and cross-portal chain resolution on top of this.

## Verification

Run locally on the branch tip:

| Check | Result |
|---|---|
| `npm run typecheck` | ✅ pass |
| `npm run build` | ✅ pass (exit 0, 341 modules, d.ts rollup clean) |
| `npm run lint` | ✅ pass (exit 0) |
| Storybook play stories | ⚠️ not run |

CI on this PR is green across `Validate`, `Validate PR title`, and `deploy-pr-preview`.

## Still to confirm

- Attribute naming (`data-ds-component`) is not yet locked — it's one of the open decisions in the instrumentation taxonomy. Easy to rename now, harder once consumers key off it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #184, opened 2026-08-12
feat(minor): emit data-ds-part and resolve interaction chains across portals

shaunrfox:
## Update — review revisions (2026-08-19)

Revised after a review pass against `INSTRUMENTATION-PLAN.md`. **Two changes reverse
behavior described earlier in this PR**, so the description above is out of date where
it disagrees with this section.

### `data-testid` stays where the consumer put it

An earlier revision moved `Select`'s `data-testid` from the combobox trigger to the
component root, so the chain scope it opens would enclose the portaled listbox. That
was a consumer break — any existing `getByTestId` against a `Select` would have
resolved to a wrapper `<div>` instead of the combobox, silently breaking `.fill()`,
`.press()` and role assertions.

**Reverted.** The test ID stays on the trigger. The scope is now opened *above* the
root instead, which decouples "which element wears the ID" from "which element opens
the scope" — only the root encloses the portal's position in the React tree, but it
doesn't need to carry the attribute to do that.

Consequence: the trigger's own `Box` opens a second scope with the same ID, so
`extendDsChain` now collapses a repeated innermost node. That also covers a consumer
wrapping a tagged DS component in `DsChainScope`. Non-adjacent repeats — a grid nested
in a grid — still contribute, because there the repeat is real structure.

### Portal roots are marked unconditionally

`data-ds-chain` is omitted when the chain is empty, which left the portal wrapper as an
anonymous `<div>`. A consumer walking up the DOM would pass straight through it into
`document.body` and build a chain from whatever it found — the silent-wrong-chain
failure this PR exists to prevent, reintroduced for any subtree nobody has tagged yet
(today: almost all of them).

Portal roots now always carry `data-ds-portal-root`. **Consumer rule:** when a DOM
walk-up reaches `data-ds-portal-root`, stop. Use `data-ds-chain` if present; otherwise
record the chain as unresolved. Never continue past a portal root.

### `data-ds-part` is not a test handle

It was introduced as the replacement query handle for the trigger. With the test ID
back on the trigger it no longer serves that purpose, and it is **not** a supported
public API — not typed, not documented, not for use in tests. Docstring rewritten to
justify it on the consumer side instead: it distinguishes interactive parts within one
component, which `data-ds-component` and the chain cannot.

### Public API surface — unchanged

`DsChainScope` remains the only new export. `data-ds-component`, `data-ds-part`,
`data-ds-chain` and `data-ds-portal-root` are all emitted automatically and are
internal.

### Verification

| Check | Result |
|---|---|
| `npm run lint` | pass (exit 0) |
| `npm run typecheck` | pass (exit 0) |
| `npm run build` | pass — 341 modules, d.ts rollup clean, ~34s |
| Storybook play stories | **10/10 pass, executed** — 7 `DsChainScope`, 3 `Select` |

The play stories were run rather than assumed. Nothing in CI executes them, so they had
never been exercised on this branch before now. Two new ones cover the revisions:
repeated-node collapse, and a findable portal boundary with an empty chain.

No release label applied — `major`/`minor`/`patch`/`release` are manual in this repo
and one of them publishes a prerelease on merge.

### Follow-up, not blocking this PR

Confirmed separately in the app repo: the legacy layout wrapper that Track B would tag
with `data-track-object` sits one level *inside* `<body>`, and `FloatingPortal` appends
to `<body>` — so portaled content lands as its sibling and can't reach it. The chain
carries `data-testid` only, so it doesn't restore that either. Tracked in
`INSTRUMENTATION-PLAN.md` v0.3; the likely fix is an app-side or `FloatingPortal`
`root` change, not a change here.

---

**Stacked on #183 — review that one first.** Base is `feat/ds-component-attribute`; it will retarget to `main` automatically once #183 lands. The diff shown here is only the 5 commits on top.

Builds the second half of the instrumentation layer: identifying *parts within* a component, and resolving interaction chains that cross a portal boundary.

## What's here

**New utils**
- `src/utils/dsPart.ts` — emits `data-ds-part` for sub-elements (triggers, panels, items).
- `src/utils/dsChain.ts` — resolves the logical parent chain across portals, so a menu item rendered in a portal still reports its owning component.

**Component changes**
- `Menu`, `SubMenu`, `Modal`, `Select`, `Tooltip` — portal roots now carry `data-ds-component`, and triggers carry `data-ds-part`.
- `Select` — test id now routes to the root element rather than an inner node.
- `Pre` and `TextInput` — stop double-spreading props. This was a real bug independent of instrumentation: props were applied twice, so the later spread silently won.

**Coverage**
- Story additions for `Select` and `TextInput`.

## Why the chain util exists

Radix-style portals render outside the component's DOM subtree, so an ancestor walk from a portaled element never reaches the component root. `dsChain` carries the association explicitly instead of inferring it from position.

## Notes

- Additive; no public API removed. `src/index.ts` exports the three new utils.
- No release label applied — per the release docs, a `minor`/`patch` label publishes a prerelease, which should be a deliberate call at merge time.

## Verification

Run locally on the branch tip (`c0c8f525`):

| Check | Result |
|---|---|
| `npm run typecheck` | ✅ pass |
| `npm run build` | ✅ pass (exit 0, 341 modules, d.ts rollup clean) |
| `npm run lint` | ✅ pass (exit 0) |
| Storybook play stories | ⚠️ not run |

### ⚠️ Note on the first `deploy-pr-preview` failure

The initial run of `deploy-pr-preview` on this PR failed, and **it was not a defect in this branch** — the build succeeded and only the final `git push` was rejected:

```
! [rejected]  gh-pages -> gh-pages (fetch first)
```

This PR and #183 were opened seconds apart, so both preview deploys raced to push to the shared `gh-pages` branch. #183 won; this one was rejected as non-fast-forward. Resolved by re-running the job.

**This exposes a real bug in `.github/workflows/deploy_gh_pages.yml`** (separate from this PR, not fixed here):

```yaml
concurrency:
  group: pages-deploy-${{ github.event.pull_request.number || 'main' }}
  cancel-in-progress: true
```

The guard is keyed **per-PR**, but the contended resource is the single shared `gh-pages` branch. Two different PRs therefore land in two different groups and deploy concurrently — precisely what the comment above it says the guard exists to prevent. Any two PRs opened close together will hit this. A constant group key (and `cancel-in-progress: false`, since cancelling a mid-push deploy leaves previews unpublished) would fix it.

## Open questions for review

- Should `dsPart` values be free-form strings or a constrained union? Currently free-form.
- The three utils are exported publicly. If they're meant to be internal-only, that should change before this merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/Autocomplete/Autocomplete.stories.tsx
+++ b/src/components/Autocomplete/Autocomplete.stories.tsx
@@ -623,3 +623,59 @@ export const TestIdReachesPortaledListbox: Story = {
   },
   parameters: { controls: { disable: true } },
 };
+
+export const DsComponentAttribute: Story = {
+  name: 'Test: data-ds-component',
+  render: () => (
+    <Box display="flex" flexDirection="column" gap="8" w="sm">
+      <Autocomplete data-testid="ds-default" aria-label="Default technology">
+        {renderOptions()}
+      </Autocomplete>
+      <Autocomplete
+        data-testid="ds-override"
+        data-ds-component="TechnologyPicker"
+        aria-label="Overridden technology"
+      >
+        {renderOptions()}
+      </Autocomplete>
+    </Box>
+  ),
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    const screen = within(canvasElement.ownerDocument.body);
+
+    // Emitted automatically on the root, without an author opting in.
+    const root = canvas.getByTestId('ds-default');
+    await expect(root).toHaveAttribute('data-ds-component', 'Autocomplete');
+
+    // The combobox input is an inner part of the root, so it stays unmarked.
+    const input = canvas.getByRole('combobox', { name: 'Default technology' });
+    await expect(input).toHaveAttribute('data-ds-part', 'trigger');
+    await expect(input).not.toHaveAttribute('data-ds-component');
+
+    // An explicitly passed value wins, still on the root and not the input.
+    const overriddenRoot = canvas.getByTestId('ds-override');
+    await expect(overriddenRoot).toHaveAttribute(
+      'data-ds-component',
+      'TechnologyPicker',
+    );
+    await expect(
+      canvas.getByRole('combobox', { name: 'Overridden technology' }),
+    ).not.toHaveAttribute('data-ds-component');
+
+    // The portaled listbox is not the Autocomplete root either.
+    await userEvent.click(input);
+
+    const listbox = await screen.findByRole('listbox');
+
+    await expect(listbox).not.toHaveAttribute(
+      'data-ds-component',
+      'Autocomplete',
+    );
+    await expect(listbox).not.toHaveAttribute(
+      'data-ds-component',
+      'TechnologyPicker',
+    );
+  },
+  parameters: { controls: { disable: true } },
+};
--- a/src/components/Autocomplete/Autocomplete.stories.tsx
+++ b/src/components/Autocomplete/Autocomplete.stories.tsx
@@ -597,15 +597,28 @@ export const TestIdReachesPortaledListbox: Story = {
     await expect(root).toContainElement(input);
     await expect(input).not.toHaveAttribute('data-testid');
 
+    // The input keeps a stable query handle through `data-ds-part`, which the
+    // component emits on its own. It marks the trigger only, never the root.
+    await expect(input).toHaveAttribute('data-ds-part', 'trigger');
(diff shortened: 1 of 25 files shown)


Etwigg (dismissed):

atkinsdavid (dismissed):
Conflictarooskis

atkinsdavid (approved):

Etwigg (approved):

Etwigg:

PR #190, opened 2026-08-21
feat(minor): stamp data-track-object on portal roots

shaunrfox:
Closes the object-name half of the portal problem that [#184](https://github.com/Cetec-ERP/Cetec-Design-System/pull/184) deferred.

## The problem

A portaled overlay mounts at `document.body`, which makes it a **sibling** of the element the application tags with the business object, not a descendant. Walking up from inside a menu, listbox or tooltip reaches `<body>` and stops. On a legacy Perl screen that means the object is never found; on a React screen, where `data-track-object` sits on individual elements, a walk-up that continues past the portal borrows whatever it lands on. Both failures are silent.

#184 solved this for the interaction chain by carrying it through React context. It could not solve it for the object, because the object is authored as markup rather than through React — a layout wrapper on legacy screens, individual elements on React screens.

## What this does

`DsChainPortalRoot` now takes the element that opened the portal and copies the value of that element's nearest `data-track-object` ancestor onto the portal root, alongside the chain it already stamps.

- `DS_OBJECT_ATTRIBUTE` and `dsObjectValue()` in `src/utils/dsChain.ts`. The walk is inclusive of the element itself and returns `undefined` when nothing is tagged, so React omits the attribute rather than emitting an empty string.
- **Nearest wins.** A legacy screen tagged on one layout wrapper and a React screen tagged on individual elements both resolve correctly from the same walk. A page with several tagged regions resolves to the region the interaction started in.
- Written **during render, never from an effect** — the same race `#184` already guards: portal content mounts late, and an effect-stamped attribute can lose to a click that lands immediately after paint.
- Five call sites pass `floating.elements.domReference`: `Autocomplete`, `Menu`, `SubMenu`, `Select`, `Tooltip`. That is the reactive element rather than `refs.domReference.current`, so nothing reads a ref during render. `Menu` was already using it.

**The collector rule is unchanged.** Stop at `data-ds-portal-root`; read `data-ds-chain` and now `data-track-object` from there. Never continue past a portal root.

## Modal passes no reference

`Modal` is driven by the `open` prop and never calls `refs.setReference`, so there is no opening element to resolve from. It emits **no object** rather than falling back to a page-level value — that fallback would be right on a legacy screen and wrong on a React screen that tags individual elements, which is exactly the silent-wrong-answer this work exists to prevent.

This is a real gap, not an oversight, and it is recorded as a passing story rather than only a code comment. If we later want object identity inside a `Modal`, the honest fix is to give `Modal` a reference or carry the object through context — not to guess here.

## Verification

Run in a real browser, not just asserted. Four new play stories in `DsChainScope.stories.tsx` cover the legacy shape, nearest-wins, an untagged page, and the `Modal` gap. Beyond those, all six portal components were exercised manually:

| Component | Resolved |
| --- | --- |
| `Select` | nearest tagged ancestor |
| `Menu` | nearest tagged ancestor — opened from **inside a Modal**, resolved the field it sits in rather than the differently-tagged page |
| `SubMenu` | nearest tagged ancestor, on both the menu and submenu roots |
| `Autocomplete` | nearest tagged ancestor |
| `Tooltip` | nearest tagged ancestor |
| `Modal` | nothing, as designed |

The Menu-inside-Modal case is the load-bearing one: nearest-wins held through a portal-in-portal with two competing tagged ancestors.

The three pre-existing portal stories from #184 still pass, so the chain work is unregressed. `lint`, `typecheck`, `check:jsdoc:strict` and `build` all exit 0.

## Context

Decided in the [Front-End Instrumentation brief](https://docs.google.com/document/d/1GuO5RymGbd8VNV5Rgb31xa2dE96zzxKPpD3H8CX4yew/edit), 2026-08-21, as option 2 of three. Option 1 — pointing the overlays at `.body-wrapper` — is cheaper than the brief assumed (that wrapper carries one `min-height` rule inside a `max-width:600px` media query and nothing else, so no containing block and no clipping context) but it only fixes legacy screens, and the 23 `only_main_content` templates have no `.body-wrapper` at all.

https://claude.ai/code/session_01LsW7ndGCBL1iKDiy8TdsoQ

### DIFF
--- a/src/components/Autocomplete/Autocomplete.tsx
+++ b/src/components/Autocomplete/Autocomplete.tsx
@@ -407,7 +407,7 @@ export const Autocomplete = (props: AutocompleteProps) => {
 
       {isOpen && !disabled && !readOnly && (
         <FloatingPortal>
-          <DsChainPortalRoot>
+          <DsChainPortalRoot reference={floating.elements.domReference}>
             <FloatingFocusManager
               context={floating.context}
               modal={false}
--- a/src/components/DsChainScope/DsChainPortalRoot.tsx
+++ b/src/components/DsChainScope/DsChainPortalRoot.tsx
@@ -1,14 +1,26 @@
 import type { ReactNode } from 'react';
 
-import { dsChainValue, useDsChain } from '~/utils/dsChain';
+import { dsChainValue, dsObjectValue, useDsChain } from '~/utils/dsChain';
 
 type DsChainPortalRootProps = {
   /** Portal contents that need DOM ancestry back to the opening subtree. */
   children?: ReactNode;
+  /**
+   * The element that opened this portal — Floating UI's `elements.domReference`
+   * at the call site. Its nearest `data-track-object` ancestor is copied onto
+   * the portal root, giving portaled content the business-object identity that
+   * DOM ancestry cannot supply.
+   *
+   * Optional because not every portal has one. A `Modal` driven by an `open`
+   * prop never sets a reference, and omitting it there yields no object rather
+   * than a guessed one.
+   */
+  reference?: Element | null;
 };
 
 /**
- * Wraps portal contents in an element stamped with the resolved chain.
+ * Wraps portal contents in an element stamped with the resolved chain and the
+ * business object the interaction belongs to.
  *
  * Internal. Render it as the outermost child of a `FloatingPortal` so the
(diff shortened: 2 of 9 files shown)


atkinsdavid (dismissed):

Etwigg (approved):

atkinsdavid (approved):

Etwigg:

